PR #17, opened 2025-08-05
OCMUI-3653 Double scrolling issue 

kelvah:
# Summary

Fixing the double scrolling issue re-introduced after HCC migrated to PF6.

# Jira

Fixes [OCMUI-3653](https://issues.redhat.com/browse/OCMUI-3653)

# Additional information

While waiting for HCC to address this on their side, I realized that we still have an issue in how we calculate our app container height. In theory this fix will make their ticket not necessary anymore.

After the PF6 upgrade the page layout changed, we can't use the viewport height and subtract header and banner size anymore. We also have new content inside the app area (the HCC breadcrumb) and some spacing below the chrome wrapper container (`.chr-render`). I had to change approach and take the size of our container and subtract the breadcrumb size.

Calculating manually the height was and remains a brittle solution. I also explored if we can remove it and rely only on flexbox. I couldn't find a quick solution mainly because of too much nesting between chrome elements and ours. I will open a ticket to investigate it properly if we agree with the fix in this PR. I think the current double scrollbar is too ugly and needs to be fixed.

`App.tsx` doesn't have any unit test coverage. I didn't add a test because I think I will end up mocking too many things inside/outside the app and Jest doesn't support DOM inspection. I can still give it a try if anyone thinks it could be valuable. 

# How to Test

- Go to https://prod.foo.redhat.com:1337/openshift
- Browse the app and verify there is no more double scrolling inside long pages
- Also verify there is no scrolling at all inside pages with little content
- Verify that drawers content scrolls independently from the rest of the page
  - Go to https://prod.foo.redhat.com:1337/openshift/overview and click one of the "Learn more" links inside one of the featured products cards
  - Go to the [ROSA wizard](https://prod.foo.redhat.com:1337/openshift/create/rosa/wizard), reach the "Account and roles" step and click on "How to associate a new AWS account"
- Try resizing the browser window and verify nothing breaks and scrolling still works as expected

# Screen Captures

|  Screenshot 2025-08-01 at 15 30 56 | Screenshot 2025-08-01 at 15 30 49 |

# Review process

## QE Reviewer

### DIFF
--- a/src/components/App/App.tsx
+++ b/src/components/App/App.tsx
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import React, { ReactNode, useState } from 'react';
+import React, { useState } from 'react';
 
 import useResizeObserver from '@react-hook/resize-observer';
 import { QueryClientProvider } from '@tanstack/react-query';
@@ -24,12 +24,6 @@ import { queryClient } from './queryClient';
 import Router from './Router';
 
 import './App.scss';
-// import patternfly styles until hcc is on PF v6
-// import '@patternfly/patternfly/patternfly.css';
-
-type Props = {
-  children?: ReactNode | undefined;
-};
 
 function getHeightUpdater(
   currentHeight: number,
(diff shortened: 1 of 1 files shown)


LiorKGOW:
Thanks for the fix @kelvah ! 
I would not block this PR on a unit test which requires lots of mocking. Although it might be better to have this covered in a unit test, mocking lots of side effects which are not related to our application might be an overkill. If you think it will take you a lot of time, maybe we could avoid it, but I am not 100% sure it is the right call to take in this case. 
I am willing to see what others think about this topic
LGTM 👍🏼

kelvah:
Hi @apecha, friendly ping on this PR waiting for your review. Thanks

apecha:
@kelvah will start looking into it later today. Didn't have bandwidth earler

apecha (approved):
Didn't found any scrolling issues, except small minor thing, where resizing windows makes at some point visible flickering double scroll bars. I have attached recording

https://github.com/user-attachments/assets/259a5739-d043-424a-8ef7-98c3b4fd23ad

For me that's not blocker for approving. I will let you decide if you want address it

lgtm

kelvah:
Follow up ticket created https://issues.redhat.com/browse/OCMUI-3741

PR #21, opened 2025-08-06
[OCMUI-3666]Fix failing e2e CI testcase

jmekkatt:
# Summary

Fixed failing e2e CI test case : osd/OSDClusterCreation.js

# Jira

Fixes [OCMUI-3666](https://issues.redhat.com/browse/OCMUI-3666)

# Additional information

The recent e2e CI workflow from the main branch reported failures on an existing test case osd/OSDClusterCreation.js.
This PR provide a fix for the same.

# How to Test

Run the following command
`yarn cypress-headless  --browser chrome --config baseUrl='<url>' --spec cypress/e2e/osd/OSDClusterCreation.js`

# Screen Captures

```javascript
DevTools listening on ws://127.0.0.1:56785/devtools/browser/2bebe329-c30c-4a78-a892-633a12d833f4
@cypress/grep: will omit filtered tests
grep and/or grepTags has eliminated all specs
Will leave all specs to run to filter at run-time

================================================================================

  (Run Starting)

  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ Cypress:        13.13.0                                                                        │
  │ Browser:        Chrome 138 (headless)                                                          │
  │ Node Version:   v22.11.0 (/usr/local/bin/node)                                                 │
  │ Specs:          1 found (OSDClusterCreation.js)                                                │
  │ Searched:       cypress/e2e/osd/OSDClusterCreation.js                                          │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘


────────────────────────────────────────────────────────────────────────────────────────────────────
                                                                                                    
  Running:  OSDClusterCreation.js                                                           (1 of 1)


  OSD cluster tests
    Create OSD cluster on AWS flow
      ✓ navigates to create OSD cluster (76264ms)
      ✓ shows an error with invalid and empty names (19529ms)
      ✓ fills OSD wizard but does not really create an OSD cluster (2564ms)

  OSD Trial cluster tests
    View Create OSD Trial cluster page
      ✓ navigates to create OSD Trial cluster and CCS is selected (1473ms)


  4 passing (2m)


  (Results)

  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ Tests:        4                                                                                │
  │ Passing:      4                                                                                │
  │ Failing:      0                                                                                │
  │ Pending:      0                                                                                │
  │ Skipped:      0                                                                                │
  │ Screenshots:  0                                                                                │
  │ Video:        false                                                                            │
  │ Duration:     1 minute, 40 seconds                                                             │
  │ Spec Ran:     OSDClusterCreation.js                                                            │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘


====================================================================================================

  (Run Finished)


       Spec                                              Tests  Passing  Failing  Pending  Skipped  
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ ✔  OSDClusterCreation.js                    01:40        4        4        -        -        - │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ✔  All specs passed!                        01:40        4        4        -        -        -  

✨  Done in 108.40s.

```

# Review process

## QE Reviewer

### DIFF
--- a/cypress/e2e/osd/OSDClusterCreation.js
+++ b/cypress/e2e/osd/OSDClusterCreation.js
@@ -17,9 +17,7 @@ describe('OSD cluster tests', { tags: ['ci'] }, () => {
     it('navigates to create OSD cluster', () => {
       cy.getByTestId('create_cluster_btn').click();
       CreateClusterPage.isCreateClusterPage();
-      cy.get('a[data-testid="osd-create-cluster-button"]', {
-        timeout: 15000,
-      }).click();
+      CreateOSDWizardPage.osdCreateClusterButton().click();
       CreateOSDWizardPage.isCreateOSDPage();
     });
 


cypress/e2e/osd/OSDClusterCreation.js
```diff
@@ -17,9 +17,9 @@ describe('OSD cluster tests', { tags: ['ci'] }, () => {
     it('navigates to create OSD cluster', () => {
       cy.getByTestId('create_cluster_btn').click();
       CreateClusterPage.isCreateClusterPage();
-      cy.get('a[data-testid="osd-create-cluster-button"]', {
-        timeout: 15000,
-      }).click();
+      cy.getByTestId('osd-create-cluster-button', { timeout: 50000 })
+        .should('not.have.class', 'pf-m-aria-disabled')
```
kelvah, line 21: I think it's better to check directly the `aria-disabled` attribute instead of the pf class. In the end the result is the same but they are meant for different purposes. In this case it feels wrong to mix them.

kelvah (approved):
Tested locally, works fine. Left only a suggestion.

cypress/e2e/osd/OSDClusterCreation.js
LiorKGOW: If it is the only possible method to select the component, that is fine, but as the guidelines suggest, we should mention that somewhere (adding a comment about it on the PR description / on the PR's code as a comment would be fine for me 😄)

In case we have a different selector we could use, that would be the approach to go for in my opinion to avoid adding attributes we could avoid to the components we use

jmekkatt: If you are referring the unittest guidelines. At the moment, we’re not following this approach in our Cypress test definitions. 🙂

LiorKGOW (approved):

LiorKGOW:
I will not block this PR on the discussion, since this PR fixes broken e2e tests on main, but let's bring this up for discussion on the next ArchSync @jmekkatt I have added it to the agenda

cypress/e2e/osd/OSDClusterCreation.js
jmekkatt: The use of `.should('not.have.attr', 'aria-disabled')` isn't working reliably in our scenario.
As an alternative, we can invoke the attribute directly, check for its absence, and then perform the click. 
For example:
```
cy.getByTestId('osd-create-cluster-button')
  .invoke('attr', 'aria-disabled')
  .should('not.exist')
  .then(() => {
    cy.getByTestId('osd-create-cluster-button').click();
  });
 ```
However, I found that relying on the class definition makes things simpler

eliranmal (commented):
thanx for this 🍪  

had [commented][1] - PTAL 



[1]: https://github.com/RedHatInsights/uhc-portal/pull/21/files#r2257496539

cypress/e2e/osd/OSDClusterCreation.js
eliranmal: IRT:

> we can invoke the attribute directly, check for its absence, and then perform the click

yes - that's what i would go with actually.   i've seen similar issues after the move to PF 6 (where aria-disabled is simply omitted) in another area

jmekkatt:
@eliranmal Hi, Could you please help with a re-review? Thanks.

cypress/e2e/osd/OSDClusterCreation.js
kelvah: > The use of .should('not.have.attr', 'aria-disabled') isn't working reliably in our scenario.

That's probably because you have to check if `aria-disabled` is undefined OR false to cover all scenarios.

> However, I found that relying on the class definition makes things simpler

True but that shouldn't be the only criteria to consider if there are alternatives. 

> oh BTW, does simply .should('be.disabled') not work for this case?

That doesn't work on `<a>` links because you can't disable them. You can trick the browser disabling pointer events but they don't support a `disabled` attribute.

eliranmal: > That doesn't work on <a> links because you can't disable them

right.. the problem makes more sense now

PR #31, opened 2025-08-07
OCMUI-3586: rosa login command update

davidaznaur:
# Summary

Update to rosa login commands to reflect new way of loginning into the ROSA CLI. Instead of offline tokens use SSO login command

# Jira

[OCMUI-3586](https://issues.redhat.com/browse/OCMUI-3586)

# Additional information

Simple string update from rosa login --env ${env} to rosa login --use-auth-code

# How to Test

Start creating an hcp cluster.
On Cluster Roles and policies step check Step 3 commands. Confirm that rosa login --use-auth-code is present.
Click on Create new OIDC config link and confirm that in the Popup rosa login --use-auth-code command is present.
Finish cluster creation BUT DO NOT add OIDC config. We need a warning on the cluster installation stepper.
Click on the link Action required on OIDC and operator roles in Cluster installation stepper. Confirm that rosa login --use-auth-code command is present.

# Screen Captures

| Screenshots of BEFORE are in the Jira ticket | image 
image
|

# Review process

## QE Reviewer

### DIFF
--- a/src/components/clusters/common/InstallProgress/ActionRequiredModal.jsx
+++ b/src/components/clusters/common/InstallProgress/ActionRequiredModal.jsx
@@ -46,7 +46,7 @@ function ActionRequiredModal({ cluster, isOpen, onClose, regionalInstance }) {
     const oidcConfigID = cluster.aws.sts?.oidc_config?.id;
     const operatorRolePrefix = cluster.aws?.sts?.operator_role_prefix;
     const installerRole = cluster.aws?.sts?.role_arn;
-    const rosaRegionLoginCommand = `rosa login --use-auth-code`;
+    const rosaRegionLoginCommand = `rosa login --use-auth-code --url ${regionalInstance?.url}`;
     const operatorRolesCliCommand = `rosa create operator-roles ${
       isHCPCluster ? '--hosted-cp' : ''
     } --prefix "${operatorRolePrefix}" --oidc-config-id "${oidcConfigID}"  --installer-role-arn ${installerRole}`;
--- a/src/components/clusters/wizards/rosa/ClusterRolesScreen/CustomerOIDCConfiguration.jsx
+++ b/src/components/clusters/wizards/rosa/ClusterRolesScreen/CustomerOIDCConfiguration.jsx
@@ -79,7 +79,11 @@ function CustomerOIDCConfiguration({
   input: { onChange },
   meta: { error, touched },
 }) {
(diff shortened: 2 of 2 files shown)


t-hendricks (approved):
LGTM!

src/components/clusters/common/InstallProgress/ActionRequiredModal.jsx
```diff
@@ -46,7 +46,7 @@ function ActionRequiredModal({ cluster, isOpen, onClose, regionalInstance }) {
     const oidcConfigID = cluster.aws.sts?.oidc_config?.id;
     const operatorRolePrefix = cluster.aws?.sts?.operator_role_prefix;
     const installerRole = cluster.aws?.sts?.role_arn;
-    const rosaRegionLoginCommand = `rosa login --url ${regionalInstance?.url}`;
+    const rosaRegionLoginCommand = `rosa login --use-auth-code`;
```
DcooperRH, line 49: I'm not sure this works for ap-southeast-1 clusters.. when using this new login command then running the next two commands, I get the error "`E: There was a problem retrieving OIDC Config '2e50baroj1tr9klnp7if3fl95u0sbs6o': OIDC Configuration '2e50baroj1tr9klnp7if3fl95u0sbs6o' not found`"

But when using the old login command, `rosa login --url https://api.aws.ap-southeast-1.stage.openshift.com`, I was then able to run the next two commands just fine and the cluster continued to install

davidaznaur: So I assume for this to work you need to be in that region. When trying to login into stage vs prod we add --env staging or production right. Probably if someone is in singapore region SSO will default for them to singapore. But that is an assumption and @kelvah mentioned that it should not be a problem. But I would double check with him.

DcooperRH: Yeah I think we may need some clarification or discussion here. I'm under the impression that the whole reason the command exists is to log into that specific region to create these oidc configs and if `rosa login --use-auth-code` doesn't do that, then I'm not sure why we would need a log in code here. I know the story says it doesn't make sense to have a mismatch in the login code on this modal compared to the rosa prerequisites page, but I think this is possibly a reason why there should be a mismatch between them. 
May need input from @kelvah or @aneelac22

DcooperRH (approved):
This command looks to be working properly with logging into the Singapore region. LGTM!

PR #59, opened 2025-08-19
OCMUI-3539: replaced install uninstall stepper in case of an error

davidaznaur:
# Summary

We have a bug that occurs during uninstallation if said uninstallation results in an error. We show installation stepper during installation, and uninstallation stepper during uninstallation. The bug is that if an error occured during uninstallation, instead of uninstallation stepper it shows installation stepper. The issue is that there is no way to differentiate when the error ocured, during installation or uninstallation. The proposal is to not show install/uninstall stepper and instead show an alert and expandable logs view for the user.

# Jira

[OCMUI-3539](https://issues.redhat.com/browse/OCMUI-3539)

# Additional information

From technical standpoint the fix is very simple. In case of installation we display install stepper, in case of uninstallation we display uninstall stepper, and in case of an error we remove stepper and display alert with a message that an error ocured during installation or uninstallation and display logs for the user to debug reasons why an error ocured.

# How to Test

1. Start installing cluster and make it fail during installation (or there is a cluster ?env=production that has an install failure on it)
2. Confirm that stepper is no longer present and alert is displayed.
3. Same thing with uninstall clluster, but I did not test this scenario, because it is not possible to "mock" uninstall failure.

# Screen Captures

| image | image |

# Review process

## QE Reviewer

- [x] _Pre-merge testing : Verified change locally in a browser (downloaded and ran code using reviewx tool)_
- [x] Updated/created Polarion test cases which were peer QE reviewed
- [x] Confirmed 'tc-approved' label was added by dev to the linked JIRA ticket

### DIFF
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/ClusterDetailsTop/ClusterDetailsTop.jsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/ClusterDetailsTop/ClusterDetailsTop.jsx
@@ -15,7 +15,6 @@ import isAssistedInstallSubscription, {
 import { HAS_USER_DISMISSED_RECOMMENDED_OPERATORS_ALERT } from '~/common/localStorageConstants';
 import { useNavigate } from '~/common/routing';
 import { normalizedProducts } from '~/common/subscriptionTypes';
-import ClusterStatusErrorDisplay from '~/components/clusters/common/ClusterStatusErrorDisplay';
 import { PreviewLabel } from '~/components/clusters/common/PreviewLabel';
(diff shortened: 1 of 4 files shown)


kdoberst (approved):
Looks good.  Tested locally by just mocking the value in the code.  One minor nit that isn't worth fixing unless there are other changes needed.

t-hendricks (approved):
Tested locally. LGTM!

kelvah (commented):
Hi @davidaznaur, I'm late but I have a couple of observations.

I see there are other alerts that are showing the error state as an installation error. See screenshot from the same kim-rosa-multi cluster in prod



Did you notice them? I assume they need to be fixed too. Do we need a separate ticket or do you want to address them here?

Also, when trying to expand the logs for the same cluster I don't see anything. Do we have a bug in `InstallationLogView`?

PR #83, opened 2025-08-27
Storybook: Rosa Wizard

dtaylor113:
# Summary

Added Storybook 'stories' for all seven steps of the Rosa wizard.  There are often several 'stories' for a particular 'step' which shows various step/screens under different conditions such as for 'Rosa Classic' .vs 'Rosa Hosted' or 'Single' .vs 'Multi' availabilities zones.

Note: there should be no code changes to any existing source or 'production' code.  This PR only contains new 'Storybook' artifacts like a new '.storybook' directory and various new '*.stories.tsx' files, or updates to existing *.stories.tsx files.

# How to Test

`yarn storybook`

# Screen Captures

CleanShot 2025-08-27 at 17 21 53

# Review process

## QE Reviewer


kelvah (changes requested):
@dtaylor113 I had a look but I don't think we should merge this.

I appreciate the effort but this looks more like an experiment. There is a considerable amount of mocking, both global and local, with no typing at all. Documenting the UI via storybook is essential but it cannot come with such high maintainability costs. Unfortunately sometimes it means having to introduce some refactoring and decouple the UI layer from the logic before being able to proceed.

This is just one of the problems, there are so many that it's difficult to address them one by one here without spending too much time. If you want to know more we can continue offline.

I also find surprising how little was explained in the PR description about the implementation considering how many changes and decisions were involved. 

As a side and more general observation, PRs this big should not be opened. Tasks like this should be broken down into smaller ones,  reviewing `19,000` changes demands too much time to do it properly. It also means that it's likely too late to correct macroscopic issues with the chosen approach.

LiorKGOW (commented):
It would take me some time to review this, I wouldn't say this PR is 'XS' 😅 
I have some concerns with merging PRs which are AI generated. They need to be THOROUGHLY reviewed since they could contain some hidden issues, or a bunch of things which don't make sense at all (even if we are talking about stories and no production code). We would still need to maintain them, and they come to test our production code, so we want to be sure they are doing what we expect them to do.

I would also include a detailed How to Test section for each step. What should we expect? I don't have much experience with Storybook, and might not know all the relevant details regarding each Wizard step. I would expect a more detailed explanation on what I should expect of those tests (aka stories added) Do they come to test edge cases ? Do we want to test the entire Wizard functionality? If so, maybe we shouldn't mock too many things, in order to see all the components working together if it is more of an e2e kind of testing. Does it include an e2e test ? 

Do you think we should break this PR down to several PRs each one addressing a single step of the Wizard? It would be much easier to review each one separately, seeing which mock relates to which step, and gradually review and merge them into our code base, if we see they are good to go

PR #88, opened 2025-09-02
OCMUI-3636: fix runtime errors in identity-providers page 'unavailable' view

LiorKGOW:
# Summary

This PR comes to address a run time error where users encountered an error thrown when they where trying to edit an Identity Provider. 
More information can be found on the 'Additional Information' section of this PR, and if you want further context you can have a look at the [linked ticket](https://issues.redhat.com/browse/OCMUI-3636) and at the [GlitchTip link](https://glitchtip.devshift.net/ocm/issues/3781371) containing logs of the incidents. 

# Jira

[OCMUI-3636](https://issues.redhat.com/browse/OCMUI-3636)

# Additional Information

This was a run time error encountered by users while they tried to update an existing IdentityProvider for an existing cluster, but for some unknown reason received an error with an error code of 401, and the cluster they were inspecting was not available. We deducted that from [the issue described on GlitchTip](https://glitchtip.devshift.net/ocm/issues/3781371) containing logs of the incidents. 
If one inspects the logs, they can find that the run time error occurred on this link `https://console.redhat.com/openshift/details/s/30E2DixHDm7owR5vRj1jXMiMYfB/edit-idp/LDAP` -> meaning when editing an existing IdentityProvider.
On the response that happened on this (11:56:38.256) time, the request to fetch the cluster failed with a status code of 401, which cause a scenario where rendered the <Unavailable /> on line 159 on <IdentityProvidersPage />. Previously the component was provided with the `cluster` object, which did not contain the expected `ErrorState` type properties (as we can see on <Unavailable />, which is a Typescript file). This was not caught by Typescript since <IdentityProvidersPage /> is a Javascript file, so Typescript couldn't have notified us about this type miss match. 

## The Fix:

The problem was a miss match between an expectation of the `Unavailable` component to receive an error via a `response` prop, but received a cluster object instead, which was derived to the `ErrorDetailsDisplay` component used by `Unavailable`. This was not caught by Typescript since the `IdentityProvidersPage` component is written in Javascript, , so Typescript couldn't have notified us about this type miss match.

# How to Test

This issue is a little hard to reproduce, but we found a way to enforce a similar situation if we hard code some of the necessary changes. I added a unit test, to create the necessary requirements for the fix to show, and also we can test it after hard coding some changes (which are explained later).
 
Generally, to test this, we need to get a situation where we receive a 401 error code when fetching a cluster (`getClusterDetails` request) in order to render the `Unavailable` component inside `<IdentityProvidersPage />`. (This would trigger the `if` on line 164 of `IdentityProvidersPage.jsx` and return the result of `errorState()` since we got a 401 error code === `<Unavailable />`).

So, in order to test that the issue was fixed locally, you need to create a cluster, which has / create an identity provider of type LDAP, go to the edit section of the identity provider and then you need to hard code the following changes: 
1. On `clusterService.ts` line 97 - Change the `clusterID` of the url to an id of a cluster which does not exist. For example, adjust the line like so: `apiRequest.get<Cluster>(`/api/clusters_mgmt/v1/clusters/${clusterID}blabla`),`
2. Enforce a status code of 401 for the `getClusterDetails` request - Hard code it on `queries/helpers.ts` line 30, to return a `401` instead of the actual status (`errorCode: error?.response?.status` -> `401`).

_Explanation:_
These changes cause `isClusterDetailsError` on line 164, provided to `<IdentityProvidersPage />`, to be `true` & we don't have a cluster, since we enforced the call to fetch one failed. This would enter the `if` on line 164, which causes the component to return the result of `errorState()` which would render `<Unavailable />`.

If we perform the necessary changes to test this PR, **without the change of this PR** (be checked out on the `main` branch and hard code the changes to test it) we would get a run time error (see the 'Before' screenshot)
Testing this PR's code **with the changes** (be checked out on this PR's branch and hard code the changes to test it), would result in an expected `<Unavailable />` component to be rendered. (see the 'After' screenshot)

# Screen Captures

Before, we had a run time type error, where the code tried to extract an undefined property from the error. You can find more details on the [GlitchTip link](https://glitchtip.devshift.net/ocm/issues/3781371) containing logs of the incidents. 

After
After - Unavailable page is presented

# Review process

## QE Reviewer

### DIFF
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/IdentityProvidersPage.test.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/IdentityProvidersPage.test.tsx
@@ -171,4 +171,37 @@ describe('<IdentityProvidersPage />', () => {
       screen.getByText('Redirected to "/openshift/details/s/mySubscriptionId#accessControl"'),
     ).toBeInTheDocument();
   });
+
+  it('displays "Unavailable" if cluster cannot be fetched when `getClusterDetails` returns a status code of 401 with no cluster available', async () => {
+    // Arrange
+    useFetchClusterDetailsMock.useFetchClusterDetails.mockReturnValue({
+      cluster: undefined,
+      isLoading: false,
+      isError: true,
+      error: { errorCode: 401, errorMessage: 'Cluster is not available', operationID: 'op123' },
+    });
+
+    // Act
+    render(<IdentityProvidersPage isEditForm />);
+
+    await waitFor(() => {
+      expect(screen.queryByRole('progressbar')).not.toBeInTheDocument();
+    });
+
+    // Assert
+    // General Unavailable message
+    expect(screen.getByText('This page is temporarily unavailable')).toBeInTheDocument();
+    expect(
+      screen.getByText(
+        'Try refreshing the page. If the problem persists, contact your organization administrator or visit our status page for known outages.',
+      ),
+    ).toBeInTheDocument();
+
+    // Error Details
+    expect(screen.getByText('Error details')).toBeInTheDocument();
+    expect(screen.getByText('Error retrieving IDP page')).toBeInTheDocument();
+    expect(screen.getByText('Error code: 401')).toBeInTheDocument();
+    expect(screen.getByText('Cluster is not available')).toBeInTheDocument();
+    expect(screen.getByText('Operation ID: op123')).toBeInTheDocument();
+  });
 });


dtaylor113 (approved):
Code LGTM, meaning the one-line code change from "response={cluster}" to "response={clusterDetailsError}".   That is a heck of a long PR description which maybe could've been summarized as "we were passing the wrong variable to the error message".   :-) 
I appreciate the unit test!  So much so, that I will rely on that passing to verify this fix.  Recreating a specific cluster, then hardcoding code changes, etc.. is too involved to test.  Hopefully QE knows of a better way to test this?  🙏

kelvah (approved):
Tested locally. The fix works as expected.

Another way to simulate the issue is by using the react query dev tools. You can make the query fetching clusters details fail using the "trigger error" button. Then you don't need the local changes anymore.

jmekkatt (approved):

PR #93, opened 2025-09-03
OCMUI-3698: Migrate Prod Build from Webpack to FEC

t-hendricks:
# Summary

This PR contains configuration changes that are needed to deploy this application utilizing FEC.

Assisted by: Cursor/Auto

# Jira

Fixes [OCMUI-3698](https://issues.redhat.com/browse/OCMUI-3698)

# Additional information

No unit tests were added for this PR, due to only having configuration changes. Also, I learned that this app do not own every item in the left-hand navigation, so I removed some items that are not relative to this repo in the `deploy/frontend.yaml` file.

# How to Test

This test requires deployment to the staging environment first. If the build and deployment to staging is successful, then it is necessary to verify that all of the services in the OCM UI is functioning in the staging environment. If the staging environment is functioning, then we proceed to build, deploy, and test in production.

# Review process

## QE Reviewer


kdoberst (approved):
Definitely needs folks who are more familiar with the build/deploy process to way in.  With that said, everything looks good to me and looks like the changes wouldn't cause an issue.

package.json
```diff
@@ -212,13 +212,13 @@
     "start": "concurrently --kill-others --prefix=name --names=SERVER,MOCK --prefix-colors=cyan,reset --passthrough-arguments 'yarn dev-server --env noproxy {@}' 'mockdata/mockserver.py' --",
     "start:fec": "concurrently --kill-others --prefix=name --names=SERVER,MOCK --prefix-colors=cyan,reset --passthrough-arguments 'HOT=true yarn dev:fec --clouddotEnv prod {@}' 'mockdata/mockserver.py' --",
     "build": "webpack $(run/progress-flag-if-tty.sh)",
-    "build:prod": "yarn build --mode=production",
-    "build:saas": ". run/sentry-env.sh && yarn build:prod --output-path 'dist'",
+    "build:prod": "MODE=prod yarn build:fec",
```
eliranmal, line 215: i feel since we now use a standard command to run fec (i.e. `build:prod`), this is essentially the turning point where we officially 'switch' - in both dev-envs and CI.

and since we're 'officially' switching, we should probably also switch up the command names to reflect that:

`command:fec` --> `command`
`command` --> `command:webpack`

eliranmal: actually that ☝️  just complicates things..

on second thought - why aren't we just removing the old commands altogether at this point?

that way we can remain with the bare command names (i.e. ditch the `:fec` suffixes)

and while we're at it - why not remove webpack.config, too?  it's not used anyway 🤔

deploy/frontend.yaml
```diff
@@ -153,21 +159,6 @@ objects:
                   title: Releases
                   href: /openshift/releases
       serviceTiles:
-        - section: ai
```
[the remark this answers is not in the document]
kelvah: I see. I think it was better to fix this config in a separate PR.

package.json
```diff
@@ -208,16 +208,16 @@
     "form-data": "^4.0.4"
   },
   "scripts": {
-    "start": "concurrently --kill-others --prefix=name --names=SERVER,MOCK --prefix-colors=cyan,reset --passthrough-arguments 'yarn dev-server --env noproxy {@}' 'mockdata/mockserver.py' --",
-    "start:fec": "concurrently --kill-others --prefix=name --names=SERVER,MOCK --prefix-colors=cyan,reset --passthrough-arguments 'HOT=true yarn dev:fec --clouddotEnv prod {@}' 'mockdata/mockserver.py' --",
+    "start:ci": "concurrently --kill-others --prefix=name --names=SERVER,MOCK --prefix-colors=cyan,reset --passthrough-arguments 'yarn dev-server --env noproxy {@}' 'mockdata/mockserver.py' --",
```
eliranmal, line 211: 1. if the intention is for this to run in CI to not break things, we also need to change the CI file/s (i think it's .github/workflows/e2e.yaml)
2. i think `start:webpack` would be better

t-hendricks: it's just a placeholder for now. the CI for `.github/workflows/e2e.yaml` will still use default start to test fec. if it fails, then the CI file can be updated to use the `start:ci`, or whatever name we give it

kelvah: We are pretty confident the e2e ci will fail 🙂 

I like `start:ci`. Its purpose is more clear and it should help to discourage people from using it when developing. It's not an alternative to regular start, we only keep it until e2e ci is "fixed".

t-hendricks: Should we continue to try testing the CI with the default `start` using `fec`? Or, just update the command in the CI to use `start:ci`?
FWIW, I don't know if it will pass or fail unless we try.

eliranmal: > the CI for .github/workflows/e2e.yaml will still use default start script to test fec

is it working then?  that test is done within the scope of this PR.


the command name isn't that important.

.github/workflows/e2e.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         with:
           cmd: install
       - name: Add hosts
-        run: make dev-env-setup
+        run: sudo yarn fec patch-etc-hosts
```
eliranmal, line 72: those are different levels of abstraction; we should change the yarn command inside `make dev-env-setup` instead

dtaylor113 (approved):
Thanks for the explanation @eliranmal, LGTM

eliranmal:
another note:  

aside from dev-env, this PR also changes the way we build the app (in jenkins PR checks verification, as well as in jenkins build for our container image that gets deployed to staging).

this, also, can only be fully verified once PR is merged, by monitoring 'staging' env deployment.

eliranmal (approved):
🍪

PR #95, opened 2025-09-03
OCMUI-3738: Prevent timing issue for next wizard step by checking RQ pending flag

zherman0:
# Summary

There is a timing issue where if you have a slower connection, the "Next" button on the wizard can enable before the error check for the version upgrade is done.  Luckily, the react-query call already has the needed variable (isPending) that I can simply check before enabling the "Next" button on the wizard.  This will ensure the query has finished.

Additionally, fixed issue where null version could be selected (ocmui-3713).

# Jira

OCMUI-3738
OCMUI-3713
 Fixes [OCMUI-3738](https://issues.redhat.com/browse/OCMUI-3738) 
Fixes [OCMUI-3713](https://issues.redhat.com/browse/OCMUI-3713) 
 
# Additional information

This is easy enough to test, but it does require setting up a cluster just the right way.

# How to Test

1. Create a WIF user if you do not have one: `ocm gcp create wif-config --name my-wif-1 --project ocm-ui-dev`
2. Go to the [OSD wizard](https://prod.foo.redhat.com:1337/openshift/create/osd)
3. Use Annual and CCS , click next
4. Use GCP with WIF account.  Select your wif account your created. Click next.
5. Add a name and then select version 4.17.37
6. Use default settings for the rest of the wizard and create the cluster.
7. Once the cluster is ready, go to the details-> settings tab
8. Now, I like to open dev tools and go to Network, then select throttling, like 3G speeds so that you can make api calls  slower.  
9. Click on the "Update" button and make sure the version 4.18.22 is selected.  The "Next" button should take a brief second (depending on your network speed) before it enables.  Click next and the acknowledgement screen should appear. If the acknowledgement screen does not appear either the bug is still present or there are no acknowledgements, but there should be if you used the versions listed before.

# Review process

## QE Reviewer

### DIFF
--- a/src/components/clusters/common/Upgrades/UpgradeWizard/VersionSelectionGrid/__tests__/VersionCard.test.tsx
+++ b/src/components/clusters/common/Upgrades/UpgradeWizard/VersionSelectionGrid/__tests__/VersionCard.test.tsx
@@ -53,10 +53,14 @@ describe('<VersionCard>', () => {
     expect(screen.queryByRole('button', { name: 'more information' })).not.toBeInTheDocument();
   });
 
-  it('should render tooltip when has unmet cluster acknowledgements', () => {
-    const newProps = { ...defaultProps, isUnMetClusterAcknowledgements: true };
+  it('should render tooltip when has unmet cluster acknowledgements', async () => {
+    const newProps = { ...defaultProps, isUnMetClusterAcknowledgements: true, isSelected: true };
 
-    render(<VersionCard {...newProps} />);
+    const { user } = render(<VersionCard {...newProps} />);
+
+    const card = getCard(newProps.version);
+    await user.click(card!);
+
     expect(screen.getByRole('button', { name: 'more information' })).toBeInTheDocument();
   });
 });


zherman0:
> Hi @zherman0, this is probably a separate bug, but I noticed some missing information in the error shown on cluster create. I was using an old WIF an got this error, notice the "WIF configs are "[]"?:

@dtaylor113 - that is the correct response.  That is the response directly from the backend API.  That specific message would be out of scope for this ticket and not UI related.

dtaylor113 (changes requested):
Hi @zherman0, I encountered an error doing the following:

1.  In Update modal, select "4.18.22", wait for Next button, click Next to see next screen.
2. Hit the Back button, select an older version, then try re-selecting "4.18.22"

At this point, I can't seem to get the Next button enabled.



image

It's strange, maybe I'm clicking too fast back-n-forth, at some point the 'Next' button doesn't seem to get enabled, no matter how long I wait.

image

Even with 'no-throttling', going 'Back' and selecting an older version takes a very long time before the Next button becomes enabled!

PR #96, opened 2025-09-03
Revert "OCMUI-3717: add NPM to renovate package managers"

eliranmal:
Reverts RedHatInsights/uhc-portal#79

as it turns out, the changes in #79 didn't affect the availability/support for NPM auto-updates, and we'd have to wait for mintmaker to support it either way (at which point no change should be necessary on our end).


dtaylor113 (approved):
LGTM

PR #106, opened 2025-09-09
OCMUI-3742 Remove deprecated wizard

kdoberst:
# Summary

This PR removes the deprecated wizard component used in the upgrade wizard modal.  Since this should not have any user changes, no unit new tests were created.

# Jira

Fixes [OCMUI-3742](https://issues.redhat.com/browse/OCMUI-3742) 

# Additional information

On the upgrade wizard when the response comes back from the back-end, it should a confirmation screen in the modal (replacing the wizard).  This was possible in the old wizard component via a "isFinishedStep" key.  This was taken away and the suggested fix is to have a conditional that hides the wizard component and shows the confirmation (aka "FinishedStep")

Here's the problem when you display the wizard as a modal - the wizard component creates the modal header.  But if we want to implement a finished step - the wizard is no longer shown which means (by default) the modal header disappears.  Of course the wizard header looks different than the standard modal header.  So as part of this PR, I created a custom modal header that mimics the wizard header - See `ModalWizardHeader`.   While it does have a custom css file, it uses the same PF variables used in the wizard header.   

# How to Test

- Create a cluster that is not at the current version.  An HCP cluster at version 18.22 makes a great example
- Ensure that the cluster is set to have "Individual updates"
- Click on Cluster details > Settings tab > Update button
- Ensure that the looks and behaves as expected.

# Screen Captures

|  Screenshot 2025-09-09 at 9 18 01 AM Screenshot 2025-09-09 at 9 18 20 AM  | Screenshot 2025-09-09 at 9 20 24 AM Screenshot 2025-09-09 at 9 21 06 AM  |

# Review process

## QE Reviewer

- [x] _Pre-merge testing : Verified change locally in a browser (downloaded and ran code using reviewx tool)_
- [x] Updated/created Polarion test cases which were peer QE reviewed
- [x] Confirmed 'tc-approved' label was added by dev to the linked JIRA ticket

### DIFF
--- a/src/components/clusters/common/Upgrades/UpgradeWizard/UpgradeWizard.tsx
+++ b/src/components/clusters/common/Upgrades/UpgradeWizard/UpgradeWizard.tsx
@@ -185,18 +185,16 @@ const UpgradeWizard = () => {
           />
         </ModalBody>
       ) : (
-        <Wizard
-          onClose={() => close()}
-          onSave={() => onSave()}
-          isVisitRequired
-          className="ocm-upgrade-wizard"
-        >
+        <Wizard onClose={() => close()} onSave={() => onSave()} className="ocm-upgrade-wizard">
           <WizardStep
             name="Select version"
             id="select-version"
             footer={{
               isNextDisabled:
-                !selectedVersion || isPostSchedulePending || isUnmetAcknowledgementsError,
+                !selectedVersion ||
+                isPostSchedulePending ||
+                isUnmetAcknowledgementsError ||
+                isUnmetAcknowledgementsPending,
             }}
           >
             {isPending ? (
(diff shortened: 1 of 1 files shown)


kdoberst:
> @kdoberst - the code looks fine. I did see some odd resizing when submitting that I can live with but thought you might want to see before I approve this. (Don't worry about the error, it's expected)
> 
>  Screencast.From.2025-09-09.13-17-40.mp4

Yea - PF really doesn't handle changing the height of the content in a modal well. Was really trying to avoid putting custom CSS in - but even in PF's examples they do put in either custom CSS or pass along a height value (which puts in custom css) 🤷   

I'll break down and put in the custom css height and fix the random broken test (dang it I missed that one)

kdoberst:
@zherman0 @t-hendricks Ready for re-review.  Added min height to the wizard so while it might bounce height still, it shouldn't be as dramatic.  Also fixed test.  Both are on a new commit so you can view just the changes.

zherman0 (approved):
LGTM.  The code seems reasonable.  There will be several merge issues with PR95 as the code in there was changed to handle bugs.  The look is good and you fixed the sizing issue on the submit screen.

jloss-redhat (approved):

PR #112, opened 2025-09-10
OCMUI-797: Changed HTTPS Proxy URL validation

meirinetta:
# Summary

In the OSD and ROSA cluster creation page, in the networking section -> cluster-wide-proxy screen, validation is successful when entering an http URL to the https URL input, there is no error message displayed and the user is able to move to the next screen once the next button is clicked.
The error message should only include the scheme prefix (https://) and the user should not be allowed to go to next screen.

# Jira

Fixes [OCMUI-797](https://issues.redhat.com/browse/OCMUI-797)

# Additional information

Fix:
The fix is in each cluster proxy page of OSD and ROSA cluster.
The issue was 'http' being part of the allowed url starts for https field. 

Unit tests:
The are two related issues to this fix - the URL error message and ability to proceed to the next step.
This PR adds tests for both aspects:
1. URL error message test - for both OSD and ROSA cluster proxy pages.
1. Next button test - for both OSD and ROSA wiizards.

# How to Test

How to reproduce:
1. Launch OSD/ROSA wizard
2. In Networking -> Configuration, choose 'Configure a cluster-wide proxy'
3. In the CLuster-wide proxy screen, enter http://www.example.com/ for the text field HTTPS proxy URL and click Next.

# Screen Captures

| Screenshot From 2025-09-10 13-39-57| Screenshot From 2025-09-10 13-38-20|

# Review process

## QE Reviewer

### DIFF
--- a/src/components/clusters/wizards/osd/CreateOsdWizardFooter.test.tsx
+++ b/src/components/clusters/wizards/osd/CreateOsdWizardFooter.test.tsx
@@ -66,7 +66,7 @@ describe('<CreateOsdWizardFooter />', () => {
     expect(screen.getByTestId(wizardPrimaryBtnTestId)).not.toHaveAttribute('aria-disabled');
   });
 
-  it("Doesn't proceed to the next step when validation fails", () => {
+  it("Doesn't proceed to the next step when validation fails", async () => {
     mockedUseFormState.mockReturnValue({
       ...useFormStateReturnValue,
       validateForm: jest.fn().mockResolvedValue({ mockError: 'Mock Error' }),
@@ -75,7 +75,7 @@ describe('<CreateOsdWizardFooter />', () => {
     const { user } = render(<CreateOsdWizardFooter {...props} />);
     const nextButton = screen.getByTestId(wizardPrimaryBtnTestId);
 
-    user.click(nextButton);
+    await user.click(nextButton);
     expect(mockOnNext).not.toHaveBeenCalled();
   });
 
--- a/src/components/clusters/wizards/osd/CreateOsdWizardFooter.test.tsx
+++ b/src/components/clusters/wizards/osd/CreateOsdWizardFooter.test.tsx
@@ -29,6 +29,7 @@ const wizardPrimaryBtnTestId = 'wizard-next-button';
 
 describe('<CreateOsdWizardFooter />', () => {
   const mockedUseFormState = useFormState as jest.Mock;
+  const mockOnNext = jest.fn();
 
   const useFormStateReturnValue = {
     isValidating: false,
@@ -39,6 +40,7 @@ describe('<CreateOsdWizardFooter />', () => {
   };
 
   const props = {
+    onNext: mockOnNext,
     onWizardContextChange: jest.fn(),
     isLoadding: false,
   };
@@ -62,5 +64,31 @@ describe('<CreateOsdWizardFooter />', () => {
     });
     render(<CreateOsdWizardFooter {...props} />);
     expect(screen.getByTestId(wizardPrimaryBtnTestId)).not.toHaveAttribute('aria-disabled');
+  });
+
+  it("Doesn't proceed to the next step when validation fails", () => {
+    mockedUseFormState.mockReturnValue({
+      ...useFormStateReturnValue,
+      validateForm: jest.fn().mockResolvedValue({ mockError: 'Mock Error' }),
+    });
+
+    const { user } = render(<CreateOsdWizardFooter {...props} />);
+    const nextButton = screen.getByTestId(wizardPrimaryBtnTestId);
+
+    user.click(nextButton);
+    expect(mockOnNext).not.toHaveBeenCalled();
+  });
+
+  it('Proceeds to the next step when validation succeeds', async () => {
+    mockedUseFormState.mockReturnValue({
+      ...useFormStateReturnValue,
+      validateForm: jest.fn().mockResolvedValue({}),
+    });
+
+    const { user } = render(<CreateOsdWizardFooter {...props} />);
+    const nextButton = screen.getByTestId(wizardPrimaryBtnTestId);
+
+    await user.click(nextButton);
+    expect(mockOnNext).toHaveBeenCalled();
   });
 });
--- a/src/components/clusters/wizards/osd/Networking/ClusterProxy.test.tsx
+++ b/src/components/clusters/wizards/osd/Networking/ClusterProxy.test.tsx
@@ -18,20 +18,21 @@ jest.mock('@patternfly/react-core', () => ({
   })),
 }));
 
-const buildTestComponent = (formValues = {}) => (
-  <Formik
-    initialValues={{
-      [FieldId.HttpProxyUrl]: '',
-      [FieldId.HttpsProxyUrl]: '',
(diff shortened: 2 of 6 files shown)


src/components/clusters/wizards/osd/Networking/ClusterProxy.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+  HTTP_PROXY_PLACEHOLDER,
+  HTTPS_PROXY_PLACEHOLDER,
+} from '~/components/clusters/common/networkingConstants';
+import { FieldId } from '~/components/clusters/wizards/osd/constants';
+import { render, screen } from '~/testUtils';
+
+import { ClusterProxy } from './ClusterProxy';
+
+jest.mock('@patternfly/react-core', () => ({
+  ...jest.requireActual('@patternfly/react-core'),
```
eliranmal, line 14: is it necessary to require the actual module?  why isn't the method mock enough for testing purposes?

src/components/clusters/wizards/osd/Networking/ClusterProxy.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+  useWizardContext: jest.fn(() => ({
+    goToStepByName: jest.fn(),
+    activeStep: { name: 'mockStepId' },
+  })),
+}));
+
+const buildTestComponent = (formValues = {}) => (
```
eliranmal, line 21: total nit:  i think we usually write these methods as 'renderSomething`.

that is, instead of calling `render(buildTestComponent())` everywhere (and since it's the only way it's being used), just change the function impl' to include `render()` and rename it to `renderTestComponent`.

src/components/clusters/wizards/osd/Networking/ClusterProxy.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  it('https URL validation fails when entering http URL', async () => {
+    const { user } = render(buildTestComponent());
+
+    const textBox = screen.getByPlaceholderText(HTTPS_PROXY_PLACEHOLDER);
+    await user.type(textBox, 'http://example.com');
+    await user.click(screen.getByPlaceholderText(HTTP_PROXY_PLACEHOLDER));
+
+    expect(
+      screen.getByText(/The URL should include the scheme prefix \(https:\/\/\)/i),
```
eliranmal, line 49: you can avoid all the headache-inducing backslack escapes, if you replace the regex with a string and state `exact: false` in the options.

(`exact` controls the matcher [precision][1], and toggles case-sensitive and sub-string matching)



[1]: https://testing-library.com/docs/queries/about#precision

src/components/clusters/wizards/osd/CreateOsdWizardFooter.test.tsx
```diff
@@ -63,4 +65,30 @@ describe('<CreateOsdWizardFooter />', () => {
...
+  it("Doesn't proceed to the next step when validation fails", () => {
+    mockedUseFormState.mockReturnValue({
+      ...useFormStateReturnValue,
+      validateForm: jest.fn().mockResolvedValue({ mockError: 'Mock Error' }),
+    });
+
+    const { user } = render(<CreateOsdWizardFooter {...props} />);
+    const nextButton = screen.getByTestId(wizardPrimaryBtnTestId);
+
+    user.click(nextButton);
```
eliranmal, line 78: the user click action should be `await`ed on!

otherwise the following assertion might yield false results (a false-negative in this case).

src/components/clusters/wizards/rosa/ClusterProxyScreen.test.jsx
```diff
@@ -94,6 +96,30 @@ describe('<ClusterProxyScreen />', () => {
     expect(screen.getByText(/Complete at least 1 of the fields above\./i)).toBeInTheDocument();
   });
 
+  it('https URL validation fails when entering http URL', async () => {
```
eliranmal, line 99: i see these tests are duplicated across the wizards.

this feels weird; it points at these tests being put at the wrong level of abstraction.

i.e. assuming both wizards reuse the same component for their cluster-wide-proxy screen, why are we not testing these flows at the common component level..?

eliranmal: (and same comment for the wizard-footer tests, which i see are also duplicated)

src/components/clusters/wizards/osd/CreateOsdWizardFooter.test.tsx
eliranmal: (and same comment for other similar instances of un`await`ed user-clicks)

eliranmal (commented):
functionality works 💯 

had some comments on the code - see inline ✏️ 

also left a couple of general comments on the main conversation page 🗳️

meirinetta:
> i'm kinda questioning the core requirement in the ticket.
> 
> why was it allowed to pass in both 'http' and 'https' for https-proxy in the first place? i'm assuming it wasn't by accident, as **http**-proxy does allow a single value (so ocam's razor; why would anyone implement a more complex validation for https-proxy unless asked).
> 
> can we verify with PM and/or OCM service team what are the actual requirements or API-spec?
> 
> [swagger docs](https://api.openshift.com/) are a bit vague on that, i'd say:
> 
> > https_proxy | string
> > HTTPSProxy is the URL of the proxy for HTTPS requests.

@eliranmal Do you mind bringing this up on the ticket please?

src/components/clusters/wizards/rosa/ClusterProxyScreen.test.jsx
meirinetta: @eliranmal Each wizard (ROSA and OSD) uses different component for the cluster-wide-proxy-screen.
They do use the same url validation function but by testing the function only, we won't be testing user functionality. (same for wizard-footer)

Let me know if you have any additional insights on this.

src/components/clusters/wizards/osd/Networking/ClusterProxy.test.tsx
meirinetta: To my understanding it is necessary, because clusterProxy relies on several functions and components.
I also saw that it was done in ClusterProxyScreen so that's why I did go on to do that.

Do you think there's a more efficient way to do it? :pray:

meirinetta: Tried removing the line and it did break, also tried to find other ways but it does become more complicated.  
I think for now I will leave it this way.

eliranmal: sure thing 👌

dtaylor113:
Responded to some of @eliranmal's questions in a comment in [OCMUI-797](https://issues.redhat.com/browse/OCMUI-797?focusedId=28075010&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-28075010)

meirinetta:
Closing PR - see comments in [OCMUI-797](https://issues.redhat.com/browse/OCMUI-797?focusedId=28075010&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-28075010)

PR #113, opened 2025-09-10
OCMUI-3683 Modal.tsx refactoring

kdoberst:
# Summary
 
The purpose of this PR is to refactor Modal.tsx to use the current version of the PF modal component.  Because there are changes to the the PF modal props, some of the props into Modal.tsx needed to change.

# Jira

Fixes [OCMUI-3683](https://issues.redhat.com/browse/OCMUI-3683) 

# Additional information

While going through the modals, there were a couple of minor bugs that were fixed.  For example the cluster name was not showing when it should or the "delete" button was not red.  These minor things were fixed.  The modal content was not touched.  The functions passed as props were not touched.

Here is a log of all the impacted modals along with before/after screenshots: https://docs.google.com/document/d/1E4e0Q8-sYGi1ewZsS5DOz28ngZKvn1AWwuCyDwa3tyo

# How to Test

Look at the list of modals in the linked document.  Spot check numerous modals to ensure that they look and work correctly.  Because it was the same changes over and over again, I don't think it is necessary to check each one to approve the code, but each one will most likely need to be checked as part of a QE review.

# Screen Captures

See:  https://docs.google.com/document/d/1E4e0Q8-sYGi1ewZsS5DOz28ngZKvn1AWwuCyDwa3tyo

# Review process

## QE Reviewer

### DIFF
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/AccessControl/ExternalAuthentication/BreakGlassCredentialNewModal.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/AccessControl/ExternalAuthentication/BreakGlassCredentialNewModal.tsx
@@ -71,6 +71,7 @@ export function BreakGlassCredentialNewModal(props: BreakGlassCredentialNewModal
           onClose={handleClose}
           modalSize="small"
           description="Add a break glass credential to access the cluster."
+          hideDefaultFooter
           footer={
             <Stack hasGutter>
               {isError && (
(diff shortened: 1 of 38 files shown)


kdoberst:
> @kdoberst, The changes look okay except that the modals are slow to update any changes, and the size of the modal also seems to be a bit large compared to the staging environment. Please take a look at the recording below for reference.

The slowness is due to API calls - which were not modified - this PR just changed the the html/css for the modal.  As far as size there was a couple of things going on - I found a couple of bugs where the size appeared to be a debugging - inline CSS (which isn't encouraged) so it was removed and/or the relative width (small, medium, large) was sent to the modal component, and this PR fixes it so the proper width was sent (aka it was a bug that the intended with was not being followed).

As far as width were there particular modals that were of concern to the point where it greatly impacts usability?

PR #137, opened 2025-09-18
OCMUI-3751 Incorrect node count on edit HCP node pool

kdoberst:
# Summary

There appears to be an error in how the minimum number of allowed nodes per machine pools for HCP clusters.  When you create an HCP cluster with exactly 2 node pools (aka 2 subnets) with a node count of 1, there is weird behavior when you go to edit the machine pools.

If the machine pools are auto scaled:
The overview tab shows an error and you cannot changes to the machine pool unless you change the node count 

If the machine pools are not autoscaled:
The number of nodes automatically changes from 1 to 2 and you can not longer choose 1.  Again - effectively you cannot save the machine pool unless you change the node count.

# Jira

Fixes [OCMUI-3751](https://issues.redhat.com/browse/OCMUI-3751) 

# Additional information

This area of the is complex so care was taken to surgically change the count only for HCP clusters and only when trying to determine the number of nodes across the cluster.  The changed method is also used to determine if the user can modify taints and/or delete a machine pool.  To keep changes isolated, a new parameter was add and only set when terminating the minimum value of nodes a machine pool can have.

Here are the rules for HCP clusters:  There can be no fewer than 2 nodes without taints across all machine pools.

# How to Test

Create (or modify) a HCP cluster so it has 2 machine pools each with autoscaling and min/max both set to 1 node
Open the edit machine pool modal and ensure there is not an error on the overview tab

Create (or modify) a HCP cluster so it has 2 machine pools each without autoscaling and number of nodes set to 1 node
Open the edit machine pool modal and the node value does not change and that 1 is an option.

# Screen Captures (AutoScale)

Screenshot 2025-09-18 at 4 21 46 PM

| Screenshot 2025-09-18 at 4 22 11 PM |  Screenshot 2025-09-18 at 4 22 41 PM |

# Screen Captures (Not autoscale)
Screenshot 2025-09-18 at 4 19 09 PM

| Screenshot 2025-09-18 at 4 20 11 PM | Screenshot 2025-09-18 at 4 20 56 PM |

# Review process

## QE Reviewer

### DIFF
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/__tests__/machinePoolsHelper.test.js
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/__tests__/machinePoolsHelper.test.js
@@ -357,6 +357,17 @@ describe('isMinimumCountWithoutTaints ', () => {
       ).toBeFalsy();
     });
 
+    it('returns true if less than 2 nodes without taints - include current machine pool', () => {
+      expect(
+        isMinimumCountWithoutTaints({
+          cluster,
+          machinePools,
+          currentMachinePoolId: 'mp-no-taints',
+          includeCurrentMachinePool: true,
+        }),
+      ).toBeTruthy();
+    });
+
     it('returns false if less than 2 autoscaled nodes without taints - no scaling', () => {
       expect(
         isMinimumCountWithoutTaints({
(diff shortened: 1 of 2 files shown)


kdoberst:
> Hi @kdoberst, how did you get the hcp cluster's machine pools to '1' min/max | node count? UI prevents entering 1's Do I need to use the cli?

@dtaylor113  You have four options.  Easiest is to use this PR because if fixes the issue ;-). The second would be to create a new cluster and at creation time you can set the machine pools to have only 1 node each.  Third option of an existing cluster is to add a third machine pool with at least 2 node pools, adjust the other 2 node pools to have only 1 node then delete the node pool with 2 nodes.  Fourth you can use an existing HCP cluster that is already set to this configuration.  I set kim-hcp-9-23-1 to not be deleted.

dtaylor113 (approved):
Verified hcp autoscale and node-count values can be set to '1' and editing them does not show any validation errors.  Code LGTM.

jmekkatt (approved):
Thanks @kdoberst for the fix. Changes looks good to me.

PR #153, opened 2025-09-24
Nested routes demo

kelvah:
# Summary

Testing out nested routes with tabs.

The tabs links still look ugly. You'll have to click on the tab name, the link is only there. We need to figure out how to combine react router `Link` with `Tab`s. We might end up having to switch to `useNavigate` or use custom CSS.

There should be no content flashing when reloading the `requests` tab directly. 

# How to Test

This is only an experiment. Run the application locally and click on cluster lists. Then you can switch between cluster list and requests, each one with a different route.


kelvah:
> Just saw this in the list. The recommend way to using tabs in RR6 is to use nested routes and use the Outlet component. After fighting with this for quite a while, I was told there is an issue with nesting routes due how HCC did their routing and there wasn't much we could do. Maybe things are now different with FEC (maybe??).

@kdoberst I see outlet is useful to decouple the parent route component from the children and avoid nesting react components. I think that's not a necessity in our case but we could eventually add it.

The historical issue we had with RR6 was that we couldn't nest routers anymore. We had our own router and we removed it when migrating from RR5 to 6.

But routes alone could still be nested. That's what I did in this PR. I don't think FEC is involved here. The demo works even by using the old webpack config.

PR #158, opened 2025-09-25
OCMUI-1500: Add security group for HCP day 1

zherman0:
# Summary

Add security groups option for day 1 (wizard) for ROSA HCP.  The only security groups that can be added are compute (worker) security groups and they will be applied globally to all pools created.

# Jira

Fixes [OCMUI-1500](https://issues.redhat.com/browse/OCMUI-1500)

# How to Test

1. Create an aws VPC and add security groups to it
2. Go to the HCP ROSA wizard
3. Select Hosted
4. Fill in values making sure you set the region to be where you created your vpc
5. On the machine pools page, select your vpc and then at the bottom should be a new "Additional secuirty groups" section.
6. Add security groups
7. Finish the cluster wizard
8. Verify on the review page the security groups are listed
9. Submit
10. Once the cluster in ready, check the details page that it list the security groups.
11. Then check the machine pools tab and verify the worker pool has the security group(s)
 
# Screen Captures

|  |  |

# Review process

## QE Reviewer


dtaylor113 (changes requested):
The location of the Security Groups in the Review step is incorrect.
Shouldn't it be under the 'default machine pool' section, because when I click on 'edit step' for MP, it navigates to the correct step which has the additional security groups defined; not the Networking step. 
However, this would just be for Rosa Hosted, as this is correct for Rosa Classic, which does define security groups in the Networking step.
image

kdoberst (approved):
Works well locally

jmekkatt:
Hi @zherman0 , 

The security group definition is not being refreshed from the wizard’s Machine Pool step.
Scenario:
When a user initially selects a VPC that has no security groups configured, and later creates new security groups in parallel through the AWS Console, these newly created groups are not reflected in the wizard. Currently, there is no option for the user to refresh or reload the security group definitions — even after using the refresh button associated with the VPC selection.
As a result, the user is forced to relaunch the ROSA wizard to populate the updated security group definitions, which is not an ideal experience.

Proposal:
We should introduce the API call
`/aws_inquiries/vpcs?fetchSecurityGroups=true
`to be triggered on every refresh action associated with the “Select VPC” refresh button.

Alternatively, we can introduce a dedicated refresh button under the Additional Security Groups section to allow users to reload the latest security group definitions without restarting the wizard.
(btw, Honestly I don't know why this wasn't considered during the story refinement process)

jmekkatt:
Hi @zherman0 , The Security Group definition displayed on the “Review and Create” page is incorrect and reflects outdated data from a previously selected VPC.

Scenario:

1. In the Machine Pool step of the wizard, the user selects a VPC and configures private subnet machine pools.
2. The user then selects Additional Security Groups associated with that VPC.
3. Later, the user changes the VPC selection by choosing a different VPC from the “Select VPC” dropdown and updates the private subnet configuration accordingly.
4. The newly selected VPC does not have any security groups associated with it.
5. When the user proceeds to the “Review and Create” page, the security groups displayed still reflect the initial VPC’s security group data, instead of being updated to match the latest VPC selection.

Expected Behavior:
The “Review and Create” page should always display security group information that corresponds to the user’s latest VPC selection, ensuring consistency with the current configuration in the wizard.

refer the recording

https://github.com/user-attachments/assets/4c699269-7853-4ec5-bab9-52402ff23d56

jmekkatt (changes requested):
Please refer the latest comments on multiple issue.

zherman0:
@jmekkatt 
1) For your first point, I already added a refresh button to the "No security groups" alert for prior work forROSA  classic,  It should have shown the refresh automatically, but there was a bug for HCP that I have now fixed. You should see the refresh in the alert (just like it appears for classic).
2) For the review page not updating the security groups when the vpc changes.  There was no code to clear the old security groups on VPC change; that was a bug that is now fixed. 

I do see other edge cases regarding deletion of security groups after they have been selected and before the cluster is created, but those kind of issues exist with lots of resources and are very difficult to protect against (i.e. someone deletes the VPC would also be an issue).

zherman0:
@jmekkatt - I was just testing the classic path and it appears the refresh button was not active for the classic path either on the wizard.  I guess you will need to test it in both HCP and classic now.  The original work for that refresh activated it only in day 2.  I thought the prior work activated in day 1, but I was mistaken.  By turning on the refresh in HCP, I have also added it to classic.

jmekkatt:
Hi @zherman0 , 
During regression testing of the OSD CCS AWS wizard, it was observed that the refresh link (when no security groups are present in the VPC) and the refresh button associated with the Security Group dropdown were both missing

Screenshot 2025-10-08 at 12 40 50 PM

Screenshot 2025-10-08 at 12 48 04 PM

PR #159, opened 2025-09-26
OCMUI-3771: Channel Group edit Button

davidaznaur:
# Summary

Group channel section on Cluster Overview page with a button allowing users to update Channel groups for the cluster.

Assisted by: Google/gemini-2.5-pro
(unit tests)

# Jira

[OCMUI-3771](https://issues.redhat.com/browse/OCMUI-3771)

# Additional information

# How to Test

Create an HCP cluster (preferably 4.16.46 version this one 100% has EUS)
After cluster was created navigate to Cluster Details page.
Locate Channel Group and confirm that Channel group is displayed.
IF logic to filter channel groups is loading, confirm that spinner is displayed.
After spinner disappears confirm that a pencil icon is displayed.
Click on the pencil icon and confirm that modal is displayed.
Try changing channel group from Stable to FAST or CANDIDATE. Confirm that error message is displayed in the modal
Try changing from Stable to Eus. Press submit. Confirm that modal closes and changes are reflected on Cluster Details page

# Screen Captures

|  | image 
image
image
|

# Review process

## QE Reviewer

### DIFF
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsLeft.jsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsLeft.jsx
@@ -22,13 +22,12 @@ import Timestamp from '../../../../common/Timestamp';
 import ClusterTypeLabel from '../../../common/ClusterTypeLabel';
 import InfrastructureModelLabel from '../../../common/InfrastructureModelLabel';
 
-import ClusterVersionInfo from './ClusterVersionInfo';
 import { ChannelGroupEdit } from './ChannelGroupEdit/ChannelGroupEdit';
+import ClusterVersionInfo from './ClusterVersionInfo';
 
 const getIdFields = (cluster, showAssistedId) => {
   let label = 'Cluster ID';
   let id = get(cluster, 'external_id', 'N/A');
-  console.log('CLUSTER', cluster);
   const assistedId = get(cluster, 'aiCluster.id', 'N/A');
   if (showAssistedId && assistedId) {
     label = `Assisted cluster ID / ${label}`;
@@ -41,7 +40,6 @@ function DetailsLeft({ cluster, cloudProviders, showAssistedId, wifConfigData })
   const cloudProviderId = cluster.cloud_provider ? cluster.cloud_provider.id : null;
   const region = cluster?.region?.id;
   const clusterID = cluster?.id;
-  console.log('DETAILS', cluster.id);
   const planType = get(cluster, 'subscription.plan.type');
   const isROSA = planType === normalizedProducts.ROSA;
   const isHypershift = isHypershiftCluster(cluster);
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsLeft.jsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsLeft.jsx
@@ -23,11 +23,12 @@ import ClusterTypeLabel from '../../../common/ClusterTypeLabel';
 import InfrastructureModelLabel from '../../../common/InfrastructureModelLabel';
 
 import ClusterVersionInfo from './ClusterVersionInfo';
+import { ChannelGroupEdit } from './ChannelGroupEdit/ChannelGroupEdit';
 
 const getIdFields = (cluster, showAssistedId) => {
   let label = 'Cluster ID';
   let id = get(cluster, 'external_id', 'N/A');
-
+  console.log('CLUSTER', cluster);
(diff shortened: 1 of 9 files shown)


kdoberst:
Ran it locally and kicked the tires a bit.  There are a couple of things.  First should be an option for all hosted clusters (ROSA, OSD, etc).  Looks like you are limiting to HCP clusters. 

Second, the modal styling looks a little off.  Adding the edit display name modal for comparison.  

Screenshot 2025-09-29 at 3 57 51 PM Screenshot 2025-09-29 at 3 58 05 PM 


Lastly, the ticket states that if the version isn't available in a given channel, that the channel should show but be disabled.  Right now I'm actually OK with that, but we would just need to add a note at the bottom of the ticket.

kdoberst (approved):
Changes look good.

shivanthi-amara (changes requested):
https://github.com/user-attachments/assets/74b42894-eafe-4a17-a2cf-38da40accb1f

Testing for this PR did not succeed in the pre-merge env. Please review the comments and the attached recording.

1)If a user does have cluster edit access, then the user clicks on the edit icon, a modal will appear allowing the user to change the channel group.

Actual Result
Instead of simply showing the Channel group, there is still a pencil icon that appears disabled. This does not seem to leave a good user experience.

Image


2) When the user creates a candidate version cluster using the CLI and tries to change the channel group, there are no options shown in the dropdown.

3) When the user creates a cluster using the UI and tries to change the channel group from stable to candidate or fast, there is no error shown on the Overview page and the details are saved successfully.

4) There is no entry in the Cluster history tab for any changes made to the version dropdown. Kindly confirm is this is out of scope for this story.

davidaznaur:
@shivanthi-amara hi, I will address 2 points straight away.
4. There was never a requirement to display any changes in Cluster history tab.
1. Pencil icon -> If you look at Delete Protection it is disabled as well. All the "editable" fields in cluster overview display pencil icon and show it disabled when edit is not available.
2. I don't know about CLI, but this is the way we "filter". There is no dedicated endpoint to get channel groups, so filtering happens the same way it happens in the wizard -> based on cluster version. If there is no candidate, nightly, fast or stable for that particular version it will not be displayed and we have no way to address that, unless there is an endpoint. In essence it is backedn issue if it is actually an issue, not the UI part

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsLeft.jsx
```diff
@@ -150,6 +152,13 @@ function DetailsLeft({ cluster, cloudProviders, showAssistedId, wifConfigData })
           </DescriptionListDescription>
         </DescriptionListGroup>
       )}
+      {useEusChannel && (
```
DcooperRH, line 155: I actually have a question on this if either of you have clarification @kdoberst @davidaznaur , I handled this feature gate similarly in my PR for the osd wizard, but I am wondering if this whole channel group edit should be behind the feature gate or only the Eus channel option?

DcooperRH: Nvm. checked with Kim and I believe how it is handled here is correct

shivanthi-amara:
@davidaznaur Tested the code changes and have some observations. please refer to the below scenario

1) Create a cluster in candidate version that has upgrade path available in the stable for example 4.19.14
2) Turn off the capability to allow non-stable-channel-group for the organization using the command
`echo '{ "key":"capability.organization.non_stable_channel_group","value":"false","internal":true }' | ocm post /api/accounts_mgmt/v1/organizations/1TBZNa7kLQTshFFKjV2AVpY4Vjm/labels`
3) On the Overview page, the channel group field shows Candidate as the selected option but once the user clicks on the pencil icon, Candidate option is not shown in the list and Stable is shown as the default one. Furthermore, the user is not able update the preferences i.e. change to Stable in this case as Save button is disabled. This leads to confusion for the user. 

https://github.com/user-attachments/assets/0a6a9bc5-4425-4686-888c-ad8fb618adbd

shivanthi-amara:
@davidaznaur I tested the changes and I am noticing a "**TypeError: Cannot read properties of undefined (reading 'channel_group')**" when I register a disconnected cluster using the UI.


Screenshot 2025-10-23 at 6 28 46 PM



Screenshot 2025-10-23 at 6 28 07 PM

https://github.com/user-attachments/assets/5afc18df-beae-4b0a-8454-1fabba2866f3

shivanthi-amara:
@davidaznaur Testing succeeded in the pre-merge environment. Kindly review the testcase and add a tc-approved label if it looks good to you.

shivanthi-amara (approved):
LGTM

PR #410, opened 2026-03-19
[OCMUI-4113, OCMUI-4282]: Enable Y-stream channel selection at day 2

t-hendricks:
# Summary

This PR replaces the Channel group property with a new Channel property in ROSA classic, HCP, and OSD clusters. The change will be in the cluster details overview tab. This feature is implemented behind the `ocmui-y-stream-channel` feature gate.

Assisted by: Cursor

# Jira

Fixes [OCMUI-4113](https://redhat.atlassian.net/browse/OCMUI-4113), [OCMUI-4282](https://redhat.atlassian.net/browse/4282)

# Additional information

This feature behaves the same as the old Channel group property, but it depends on different data from the backend. This PR was developed on top of PR #373, so it depends on it to be merged in first and I will rebase this PR after. All of the commits associated with OCMUI-4113 focuses on this PR. This PR will also fix the version types workarounds.

# How to Test

1. Verify that [`ocmui-y-stream-channel`](https://ocm-stage.unleash.devshift.net/projects/default/features/ocmui-y-stream-channel) feature gate is active, if it's not already
2. Create a new ROSA HCP cluster with a version < 4.21.5
3. Once the cluster is installed; verify that the "Channel group" property is replaced with "Channel"
4. Click on the blue pencil, change the channel value, then save
5. Open your browser's dev-tools > network tab and find your cluster id in the list of requests
6. Search through the response, locate the "channel" property, verify that the value matches the channel in the UI
7. Repeat steps 2 - 6 for OSD clusters

# Screen Captures

| Screenshot 2026-03-18 at 11 33 35 PM | Screenshot 2026-03-18 at 11 34 05 PM |
| Screenshot 2026-03-18 at 11 36 19 PM | Screenshot 2026-03-18 at 11 37 16 PM |

# Review process

## QE Reviewer

[OCMUI-4113]: https://redhat.atlassian.net/browse/OCMUI-4113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OCMUI-4282]: https://redhat.atlassian.net/browse/OCMUI-4282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
```diff
@@ -0,0 +1,186 @@
...
+  clusterID: string;
+  channel: string;
+  cluster: CanEditCluster;
+};
+
+export interface CanEditCluster extends Cluster {
+  canEdit: boolean;
+}
+
+const ChannelEditModal = ({
```
zherman0, line 56: Nit: Can you rename this file to ChannelEditModal to match the component name?

src/queries/ChannelEditQueries/useMutateChannel.ts
```diff
@@ -0,0 +1,29 @@
+import { useMutation } from '@tanstack/react-query';
+
+import { clusterService } from '~/services';
+
+import { formatErrorData } from '../helpers';
+
+export const useMutateChannel = () => {
```
zherman0: Total Nit here (sorry): I know you were just copying the way useMutateChannelGroup did things, but the term "mutate" does not really fit our general terminology.  I would prefer this be called useEditChannel or useUpdateChannel... maybe something even more detailed like useEditChannelOnCluster since it uses the editCluster service.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/useGetChannelsData.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const topLevelChannels = (versionsData as { available_channels?: string[] })
+      ?.available_channels;
+    if (topLevelChannels?.length) {
+      return topLevelChannels.map((channel) => ({ value: channel, label: channel }));
+    }
+    const matchingVersion = rawId
+      ? items.find((el: VersionWithChannels) => el.raw_id === rawId)
+      : items[0];
+    const channels = matchingVersion?.available_channels ?? [];
+    return channels.map((channel) => ({ value: channel, label: channel }));
```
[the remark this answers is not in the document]
t-hendricks: @coderabbitai it seems that this suggestion is outdated. Can you verify if this is still an issue?

src/components/clusters/wizards/osd/ClusterSettings/Details/Details.tsx
```diff
@@ -247,19 +249,28 @@ function Details() {
...
   };
 
-  const availableVersions = getInstallableVersionsResponse.versions.filter(
-    (version: Version) => version.channel_group === channelGroup,
-  );
+  const availableVersions = isYStreamChannelEnabled
+    ? getInstallableVersionsResponse.versions
+    : getInstallableVersionsResponse.versions.filter(
+        (version: Version) => version.channel_group === channelGroup,
+      );
```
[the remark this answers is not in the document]
t-hendricks: The purpose of this change was to differentiate the logic between two different features. Therefore, if the Y-stream feature flag is enabled, then we do not care about the channel group filter. Check for context and verify if this is still an issue.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/useGetChannelsData.tsx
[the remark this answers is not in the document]
t-hendricks: If the matching version can not be found with raw_id, then why set a version?

src/components/clusters/wizards/osd/ClusterSettings/Details/Details.tsx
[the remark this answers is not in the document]
t-hendricks: @coderabbit, if this is no longer an issue, then resolve.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
t-hendricks: Renaming this file would suggest the default export is a reusable modal that could be used anywhere. That would be misleading. `ChannelEdit` is the full feature that contains the `ChannelEditModal`.

src/queries/ChannelEditQueries/useMutateChannel.ts
t-hendricks: Fixed in the latest commit.

src/components/clusters/ClusterDetailsMultiRegion/clusterDetailsHelper.ts
```diff
@@ -105,6 +105,8 @@ const formatChannelGroupName = (channelGroup: string) => {
   return channelGroup ? channelGroup.charAt(0).toUpperCase() + channelGroup.slice(1) : 'N/A';
 };
 
+const formatChannelName = (channel: string) => (channel && channel.trim() !== '' ? channel : 'N/A');
```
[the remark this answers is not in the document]
t-hendricks: In this case, display whatever the backend returns. The check is simply for channels that may contain only whitespace.

zherman0 (approved):
LGTM.  Tested locally and reviewed the code.  

Update: I ended up having some concerns on types and the look of the modal with how it matches the other edit modals on the overview tab.  I decided to address those in the followup PR440 (OCMUI-4114).

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
```diff
@@ -0,0 +1,186 @@
...
+  optionsDropdownData: {
+    value: string;
+    label: string;
+  }[];
+};
+
+type ChannelEditProps = {
+  clusterID: string;
+  channel: string;
+  cluster: CanEditCluster;
```
lizagilman: The correct type to use is `AugmentedCluster`, which already includes the `canEdit` property and the `canUpdateClusterResource` property which is the permission we should check here.
The Cluster type imported in this file represents the raw data we receive from the API, but it’s not the actual shape we use throughout the cluster details. We normalize and extend this data (for example, adding permission fields like `canEdit`, `canDelete`, etc.).
This should also be fixed in `ChannelGroupEdit.tsx`.

`canEdit` checks whether the user has permission to update the subscription (resource).
`canUpdateClusterResource` checks whether the user has permission to update the cluster resource.
Since the backend for the endpoint we call here validates permissions based on updating the cluster resource, we should align with that check here.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
```diff
@@ -0,0 +1,186 @@
...
+          },
+        );
+      }}
+    >
+      {(formik) => (
+        <Modal
+          id="edit-channel-modal"
+          title="Edit channel"
+          variant={ModalVariant.small}
+          onClose={handleClose}
```
lizagilman: nit: you can use `onClose` directly . (This could also be applied in `ChannelGroupEdit.tsx`, though it’s really just a small nitpick.)

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
```diff
@@ -0,0 +1,186 @@
...
+}
+
+const ChannelEditModal = ({
+  clusterID,
+  isOpen,
+  onClose,
+  channel,
+  optionsDropdownData,
+}: ChannelEditModalProps) => {
+  const { mutate, isError, error, isPending } = useEditChannelOnCluster();
```
lizagilman, line 63: Regarding `isPending` - After the user clicks “Save,” there should be some indication of a loading state. Typically, when submitting from a modal, ui shows a spinner on the button or replace the modal body content with a loading state.

lizagilman: Another thought - we should keep this consistent across the app. @kelvah, do we have a standard pattern for this?

kelvah: I think we are not yet consistent on this.

Showing the submit button in loading state and disabling it is enough. No need to replace the modal body with a spinner. ~~One loader is sufficient~~.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/useGetChannelsData.tsx
```diff
@@ -0,0 +1,50 @@
...
+    cluster.billing_model === SubscriptionCommonFieldsClusterBillingModel.marketplace_gcp;
+  const isWIF = cluster.gcp?.authentication?.id === GCPAuthType.WorkloadIdentityFederation;
+  const organization = useGlobalState((state) => state.userProfile.organization.details);
+  const unstableOCPVersionsEnabled =
+    useFeatureGate(UNSTABLE_CLUSTER_VERSIONS) && hasUnstableVersionsCapability(organization);
+
+  const { data, isLoading } = useFetchInstallableVersions({
+    isRosa,
+    isMarketplaceGcp,
+    isWIF,
```
kelvah: Given we already know the version of the cluster we are editing, you can fetch just that version and simplify this part:

`/api/clusters_mgmt/v1/versions/{version_id}`

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
```diff
@@ -0,0 +1,186 @@
...
+            hint={
+              <>
+                {constants.channelHint}{' '}
+                <ExternalLink href={docLinks.OCP_UPDATE_CHANNELS}>Learn more</ExternalLink>
+              </>
+            }
+          />
+        </DescriptionListTerm>
+        <DescriptionListDescription>
+          {formatChannelName(channel ?? '')}
```
kelvah: A channel could be just an empty string? Also we need to format them?

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
lizagilman: I was also thinking button in loading state is the better choice :+1:  Anyway I didn't imply for 2 spinners.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
t-hendricks: Resolved in commit: [4430d68](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/4430d685a60f2132d4873a865603b1d21d1acdc9)

Thanks for pointing this out. I wasn't aware of this pattern. Using this type could eliminate the need of the `useGetChannelsData` hook, due to `AugmentedCluster` including the `Version` type, which contains the `available_channels` property for a particular cluster. 

I'm thinking about removing the hook entirely and just rely on your mentioned cluster type. WDYT?

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
t-hendricks: Resolved in commit: [493499f](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/493499f131deecbb229fe5d0db764f1d977d3230)

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
t-hendricks: Resolved in commit: [bd8f96e](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/bd8f96e8faef4bf54dee575e0dcc569c50b68e8c)

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
t-hendricks: I was thinking of customers that may have old clusters that may not have any channel data. Or if they have access to a version that may not have any available channels, e.g. nightly versions. The `formatChannelName` method is misleading. It's not really formatting the string. It's choosing what to show in the UI: the API value or 'N/A'. I can change the name for something more meaningful, such as `getChannelDisplayValue`. WDYT?

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/useGetChannelsData.tsx
t-hendricks: It may not be necessary to fetch versions anymore, due to the `Version` type included with the cluster details. I left a [reply](https://github.com/RedHatInsights/uhc-portal/pull/410#discussion_r3073700355) to Liza's comment explaining my next strategy.

kelvah: Even better.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
kelvah: A channel will be assigned to all existing clusters before the feature goes GA.

If you still want to be careful I guess you can do an inline `?? "N/A"`. Should be enough.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
lizagilman: I didn't follow much on the data flow here but yes if this is the same piece of data needed here then yes why not

t-hendricks: Resolved in commit: [1c53ac1](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/1c53ac1d12bc61fca4e7949fbac9b738068ab2b3)

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/useGetChannelsData.tsx
t-hendricks: Resolved in commits: [4430d68](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/4430d685a60f2132d4873a865603b1d21d1acdc9) && [1c53ac1](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/1c53ac1d12bc61fca4e7949fbac9b738068ab2b3)

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
t-hendricks: Resolved in commit: [73388b6](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/73388b6df976c48051d7d6d25973151bee6976f5)

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
kelvah: This thread makes me think It's probably time to start converting cluster details to TS. We are doing quite some work around the overview tab. We are going to also work on machine pools soon. It's difficult write new components when the rest is still JS. I'll start creating a ticket.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelGroupEdit/useGetChannelGroupsData.tsx
```diff
@@ -9,9 +9,13 @@ import { UNSTABLE_CLUSTER_VERSIONS } from '~/queries/featureGates/featureConstan
 import { useFeatureGate } from '~/queries/featureGates/useFetchFeatureGate';
 import { useGlobalState } from '~/redux/hooks';
 import { SubscriptionCommonFieldsCluster_billing_model as SubscriptionCommonFieldsClusterBillingModel } from '~/types/accounts_mgmt.v1';
-import { Cluster } from '~/types/clusters_mgmt.v1';
+import type { Cluster } from '~/types/clusters_mgmt.v1';
+import type { AugmentedCluster } from '~/types/types';
 
-export const useGetChannelGroupsData = (cluster: Cluster, canEdit: boolean) => {
+export const useGetChannelGroupsData = (
+  cluster: AugmentedCluster & Partial<Pick<Cluster, 'billing_model'>>,
```
lizagilman, line 16: You can access `cluster.subscription.cluster_billing_mode` and then you don't need this workaround with `Pick`. 

There's a much bigger type issue here - managed clusters being narrowed to FakeCluster, and losing fields from API. But that's beyond the scope of this PR. cc @kelvah

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+    };
+    render(
+      <ChannelEdit clusterID="cluster-123" channel="stable-4.16" cluster={clusterNoChannels} />,
+    );
+
+    expect(screen.getByText('stable-4.16')).toBeInTheDocument();
+    expect(screen.queryByTestId('channelModal')).not.toBeInTheDocument();
+  });
+
+  it('should not render an edit button if cluster cannot update cluster resource', () => {
```
lizagilman:
```suggestion
  it('should not render an edit button if user cannot update cluster resource', () => {
```

t-hendricks: Resolved in commit: [63f84b7](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/63f84b76e168d12a59a4fee7f49503dd9800ea5d)

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelGroupEdit/useGetChannelGroupsData.tsx
t-hendricks: IMHO, diving too deep in this component is unnecessary. The new Channel field eventually will replace the Channel Group field, so this component and any related logic will end up as dead code.

lizagilman: @kelvah I can open a ticket about this tech debt if we don't have one already

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
```diff
@@ -0,0 +1,179 @@
...
+          />
+        </DescriptionListTerm>
+        <DescriptionListDescription>
+          {channel ?? 'N/A'}
+          {canUpdateClusterResource && hasChannelOptions ? (
+            <EditButton
+              data-testid="channelModal"
+              ariaLabel="Edit channel"
+              onClick={() => setIsModalOpen(true)}
+              isAriaDisabled={!canUpdateClusterResource || !isClusterReady}
```
lizagilman: nitpick:

```suggestion
              isAriaDisabled={!isClusterReady}
```
Since L167 already checks for `canUpdateClusterResource`

lizagilman (approved):

t-hendricks:
Hey @zherman0, a lot of code has changed since your last review. Can you review the latest changes and share any feedback? There shouldn't be any changes to the UI/UX. Most of the changes is related to types.

lizagilman:
@t-hendricks  Should this feature be hidden for non managed clusters?

t-hendricks:
> @t-hendricks Should this feature be hidden for non managed clusters?

@lizagilman,
I'm not sure. It's not mentioned in the Jira epic. I don't think we're currently hiding the Channel Group field for non managed clusters in stage/prod.

I'm assuming this feature should reflect that.

@kelvah WDYT?

lizagilman:
> > @t-hendricks Should this feature be hidden for non managed clusters?
> 
> @lizagilman, I'm not sure. It's not mentioned in the Jira epic. I don't think we're currently hiding the Channel Group field for non managed clusters in stage/prod.
> 
> I'm assuming this feature should reflect that.
> 
> @kelvah WDYT?

Could you please double-check with PM? If this isn’t supported for non-managed, it would make sense to hide it. Still, it’s best to confirm with PM.

t-hendricks:
> Could you please double-check with PM? If this isn’t supported for non-managed, it would make sense to hide it. Still, it’s best to confirm with PM.

I received confirmation from the PM to hide the channel from non-managed clusters. Resolved in commit: [ac9faf6](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/ac9faf6520483e8169fad201c1216b5a9fa99fe0)

zherman0 (approved):
LGTM.  Re tested clusters on Day 2 and reviewed code.  All looks good.

shivanthi-amara:
@t-hendricks Consider the below scenario where there are no channels available on the Overview page.

**Empty / unavailable channels**

If available_channels is missing or empty for the matching version (or there is no match), the edit button is not shown

The row still shows the current channel value (or N/A when absent)


For this scenario, I created a candidate cluster using the CLI as shown below. Once the cluster is ready as you can see in the recording there are no other channels available. In this case edit button should be disabled but it remains enabled. 


`rosa create cluster --cluster-name rosa4220ec5 --sts --role-arn arn:aws:iam::676673797948:role/ManagedOpenShift-HCP-ROSA-Installer-Role --support-role-arn arn:aws:iam::676673797948:role/ManagedOpenShift-HCP-ROSA-Support-Role --worker-iam-role arn:aws:iam::676673797948:role/ManagedOpenShift-HCP-ROSA-Worker-Role --operator-roles-prefix rosa4220ec5-m8u3 --oidc-config-id 24omf0c5mmceod3frfpvlc0nlv6cgb6p --region us-west-2 --channel-group candidate --version 4.22.0-ec.5 --ec2-metadata-http-tokens optional --replicas 2 --compute-machine-type m5.xlarge --machine-cidr 10.0.0.0/16 --service-cidr 172.30.0.0/16 --pod-cidr 10.128.0.0/14 --host-prefix 23 --subnet-ids subnet-0238a9a0bb77d9cd7,subnet-071524b57b7d8cb46 --hosted-cp --additional-compute-security-group-ids sg-0529211c6eba2118d,sg-0ba7cb5be6fd2dd95 --billing-account 676673797948
I: To view a list of clusters and their status, run 'rosa list clusters'
I: Cluster 'rosa4220ec5' has been created.
I: Once the cluster is installed you will need to add an Identity Provider before you can login into the cluster. See 'rosa create idp --help' for more information.
`

https://github.com/user-attachments/assets/84290174-8e08-420a-924d-a33252fefa73

t-hendricks:
> @t-hendricks Consider the below scenario where there are no channels available on the Overview page.
> 
> **Empty / unavailable channels**
> 
> ......

@shivanthi-amara, good catch! Thanks for bringing this up. It should be fixed in this commit: [3b6fd43](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/3b6fd431f057cf1e80ba5b709156ca9881c66e0e)

shivanthi-amara:
> > @t-hendricks Consider the below scenario where there are no channels available on the Overview page.
> > **Empty / unavailable channels**
> > ......
> 
> @shivanthi-amara, good catch! Thanks for bringing this up. It should be fixed in this commit: [3b6fd43](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/3b6fd431f057cf1e80ba5b709156ca9881c66e0e)

@t-hendricks sure, let me retest the PR and get back to you with updates.

shivanthi-amara (changes requested):
@t-hendricks thanks for the changes.

Consider the below scenario where the user is trying to perform a cluster upgrade. As per the information available on the epic, 

> 
> It will not be possible to upgrade from Y to Y+1 without changing channel first (from stable-4.15 to stable-4.16 for example). Suggested upgrades and the upgrade process stay unchanged.
> 
> Given that in order to do Y upgrades it will be necessary to switch channels first, we want to provide a way to change channels in the “Cluster > Settings” tab too. We will add another card before the “Update status” one with the channel switching ability, similarly to what we do in the “Cluster > Overview” tab.

I created a HCP cluster in EUS channel with version 4.16.34 keeping the other options default.

The current behavior of the cluster upgrade process presents some issues:
**Channel Switching Failure**: After the cluster was ready, an attempt to switch the upgrade channel to stable/fast resulted in a backend error.
**Incorrect User Experience**: Users are shown a list of channel options but are prevented from selecting a desired version. This is confusing, as the system should allow channel changes and then display the appropriate list of available upgrades for that new channel.
**Version Availability Discrepancy**: When attempting to switch to the eus-4.18 channel, the upgrade list correctly showed 4.17 versions as available. However, the system prevented the user from actually switching to the stable 4.17 channel.

Image

Image



Kindly investigate this behavior. Attached my findings here.

https://github.com/user-attachments/assets/d40d3129-98c2-4ce2-a544-0617d5854287

https://github.com/user-attachments/assets/956413c4-fe07-4997-aeb9-9a5ccc6a74ff

t-hendricks:
> @t-hendricks thanks for the changes.
> 
> Consider the below scenario where the user is trying to perform a cluster upgrade. As per the information available on the epic,
> 
> > ......

@shivanthi-amara, thank you for bringing this up. I shared this observation with the backend to investigate further. Once I receive any resolution, I'll update here.

shivanthi-amara:
> > @t-hendricks thanks for the changes.
> > Consider the below scenario where the user is trying to perform a cluster upgrade. As per the information available on the epic,
> > > ......
> 
> @shivanthi-amara, thank you for bringing this up. I shared this observation with the backend to investigate further. Once I receive any resolution, I'll update here.

@t-hendricks sounds good, I will continue with my other test scenarios and share my update here. The earlier issue with edit button not shown in case of empty available_channels is resolved now.

shivanthi-amara:
@t-hendricks I can confirm that all the other test scenarios work fine. just adding our conversation update here on the feature gate testing. 
**Scenario**
When the feature flag ocmui-unstable-cluster-versions and the other flag ocmui-allow-eus-channel is turned off, The only option shown to the user in the channel dropdown is Stable. However the unstable channel types and eus channels are also displayed as shown in the screenshot. Don’t we have a way to turn off the unstable cluster versions like fast and candidate? - Trevor trying to figure out if we need to set that feature flag, or just rely on the backend's feature flag.

shivanthi-amara:
@t-hendricks tested the code changes with unstable-versions-feature-flag. I am unable to view the fast, candidate and other channels inspite of turning the feature flag on/off. Kindly take a look

t-hendricks:
> @t-hendricks tested the code changes with unstable-versions-feature-flag. I am unable to view the fast, candidate and other channels inspite of turning the feature flag on/off. Kindly take a look

@shivanthi-amara, I don't see any recordings or screenshots. Did you pull the last commit?

t-hendricks:
Hey @jmekkatt, would you mind reviewing this PR? I believe Lakshmi will be out until next week.

jmekkatt:
Hi @t-hendricks , I don’t see any specific test cases linked to this feature, so I'm unclear about the current QE coverage. I have reviewed the PR threads (which are quite large) to get some context. That said, I will proceed with the review based on the acceptance criteria until Lakshmi is back. Thanks

jmekkatt (commented):
Hi @t-hendricks , Updating the channel via "Edit channel" is always blocked with an "unsupported upgrade policy" error when the cluster is set to Recurring updates. 
If changing the channel isn't supported in this mode, we should disable the edit option and show an appropriate tooltip instead of surfacing a backend error
Review the recording 

https://github.com/user-attachments/assets/d3993c45-4f59-4e57-8f24-3bd2d733cc53

jmekkatt:
@t-hendricks , After changing the channel via the "Edit channel" dialog on the cluster details page and clicking Save, the updated channel value is not reflected immediately. The UI continues to show the old channel value until a background API refetch completes, which can take several seconds (or longer on slow networks).

**Steps**
1. Navigate to the cluster details page for a managed, ready cluster
2. In the Overview section, locate the Channel row 
3. Click the edit (pencil) icon next to the channel value
4. The "Edit channel" modal opens with the current channel pre-selected
5. Select a different channel from the dropdown 
6. Click the Save button
7. The modal closes
8. Immediately observe the Channel value shown on the cluster details page

**Current Result**
The modal closes immediately after the PATCH API call succeeds
The Channel row still shows the old value  for several seconds
There is no loading indicator (spinner, skeleton, or any visual cue) on the Channel row to indicate that an update is in progress
After a delay (typically 2-5 seconds, longer on slow networks), the value eventually updates.

**Expected Result**
After clicking Save and the modal closing, the Channel value on the cluster details page should immediately reflect the new value 
OR, at minimum, a loading indicator (spinner/skeleton) should appear on the Channel row to indicate the value is being refreshed, so the user knows the save was successful and an update is in progress
 

https://github.com/user-attachments/assets/02c03a30-67fc-4a26-89fe-ec667cb4551c

jmekkatt:
@t-hendricks Hi,  Unstable Channels (candidate, fast) shown without feature flag enabled
**feature flags settings**
ocmui-unstable-cluster-versions  : disabled
ocmui-y-stream-channel : enabled

**Steps** 
1. Navigate to a managed, ready cluster.
2. Click the edit pencil next to "Channel"
3. Observe the dropdown options

**Current Result**
All channels including candidate-X.Y, fast-X.Y etc are shown in the dropdown regardless of feature gate state.

**Expected Result**
Unstable channels should be filtered out unless ocmui-unstable-cluster-versions feature flag is enabled

https://github.com/user-attachments/assets/d3bd827d-de46-4643-b261-54b509584f56

jmekkatt:
Hi @t-hendricks 
An org member with the Machine Pool Editor role is currently able to change the cluster channel, which seems incorrect.

Pre-condition
Open a cluster (ROSA/OSD)
Go to Access Control → OCM Roles and Access
Grant the Machine Pool Editor role to an org member user

Steps to reproduce
Log in as the org member user
Open the cluster
Click the edit (pencil) icon next to Channel
Change the channel value and save

Current result
An org member with the Machine Pool Editor role is able to modify the channel.

Expected result
An org member with the Machine Pool Editor role should not be allowed to modify the channel.

lizagilman:
> Hi @t-hendricks An org member with the Machine Pool Editor role is currently able to change the cluster channel, which seems incorrect.
> 
> Pre-condition Open a cluster (ROSA/OSD) Go to Access Control → OCM Roles and Access Grant the Machine Pool Editor role to an org member user
> 
> Steps to reproduce Log in as the org member user Open the cluster Click the edit (pencil) icon next to Channel Change the channel value and save
> 
> Current result An org member with the Machine Pool Editor role is able to modify the channel.
> 
> Expected result An org member with the Machine Pool Editor role should not be allowed to modify the channel.

If the request is actually successful (no backend error returned), this implies it's a backend issue.

jmekkatt:
> > Hi @t-hendricks An org member with the Machine Pool Editor role is currently able to change the cluster channel, which seems incorrect.
> > Pre-condition Open a cluster (ROSA/OSD) Go to Access Control → OCM Roles and Access Grant the Machine Pool Editor role to an org member user
> > Steps to reproduce Log in as the org member user Open the cluster Click the edit (pencil) icon next to Channel Change the channel value and save
> > Current result An org member with the Machine Pool Editor role is able to modify the channel.
> > Expected result An org member with the Machine Pool Editor role should not be allowed to modify the channel.
> 
> If the request is actually successful (no backend error returned), this implies it's a backend issue.



> > Hi @t-hendricks An org member with the Machine Pool Editor role is currently able to change the cluster channel, which seems incorrect.
> > Pre-condition Open a cluster (ROSA/OSD) Go to Access Control → OCM Roles and Access Grant the Machine Pool Editor role to an org member user
> > Steps to reproduce Log in as the org member user Open the cluster Click the edit (pencil) icon next to Channel Change the channel value and save
> > Current result An org member with the Machine Pool Editor role is able to modify the channel.
> > Expected result An org member with the Machine Pool Editor role should not be allowed to modify the channel.
> 
> If the request is actually successful (no backend error returned), this implies it's a backend issue.

Just to clarify, the UI expectation is to disable the pencil icon for the channel field and show a tooltip, similar to the delete protection feature
Screenshot 2026-04-28 at 5 57 59 PM

t-hendricks:
Hey @jmekkatt, thank you for taking the time to review this PR on short notice. I added couple of commits to solve your recent observations. I'll address the other comments below.

- > ....
If changing the channel isn't supported in this mode, we should disable the edit option and show an appropriate tooltip instead of surfacing a backend error

  - Resolved in commit: [f835a49](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/f835a49923bde2777809b64b981a9f24db2323b6)
  
- > ....
After clicking Save and the modal closing, the Channel value on the cluster details page should immediately reflect the new value
OR, at minimum, a loading indicator (spinner/skeleton) should appear on the Channel row to indicate the value is being refreshed, so the user knows the save was successful and an update is in progress

  - Resolved in commit: [f8cf56a](https://github.com/RedHatInsights/uhc-portal/pull/410/commits/f8cf56a8dd71fb67587c82387be0c779676d1ac2)
  
- > ....
Unstable channels should be filtered out unless ocmui-unstable-cluster-versions feature flag is enabled

  - The TLDR; we're trying to avoid adding any filters within client side code and just display the returned data from the API. The backend also has the ability to gate unstable channels. Rather than testing with a feature flag on the client-side, if possible, we should try testing these features with the capabilities removed. If you have more questions, dm me in slack, so we can avoid large threads.
  
- > ....
Just to clarify, the UI expectation is to disable the pencil icon for the channel field and show a tooltip, similar to the delete protection feature

  - I think Liza is right. The delete protection feature is also _not_ disabled when a user is granted the _**Machine pool editor**_ role. However, the backend does prevent me from enabling the protection. Whereas the channel field does not. Review the recording:
  

https://github.com/user-attachments/assets/2e9619ab-6eb3-4b3d-b696-106a7fbfdaed

lizagilman:
Looks like the delete protection being enabled is a UI issue (Jaya/ Trevor please open a ticket :pray: ) and the channel field is a backend issue. 
There has been backend permission issues with this role in the past but I was under impression it's been resovled.
 JIRA: [[OCM] MachinePoolEditor has more permissions besides machinepool resource](https://redhat.atlassian.net/browse/OCM-1503)

t-hendricks:
> Looks like the delete protection being enabled is a UI issue (Jaya/ Trevor please open a ticket 🙏 ) and the channel field is a backend issue. There has been backend permission issues with this role in the past but I was under impression it's been resovled. JIRA: [[OCM] MachinePoolEditor has more permissions besides machinepool resource](https://redhat.atlassian.net/browse/OCM-1503)

@lizagilman, this may be a bigger issue with the `AugmentedCluster` type and `canUpdateClusterResource`. The edit pencil button is enabled/disabled based on that property for both [delete protection](https://github.com/RedHatInsights/uhc-portal/blob/main/src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx#L151) and [channel](https://github.com/RedHatInsights/uhc-portal/pull/410/changes#diff-c97e891bdbe7faf82e14dff2ab62b28622ce0db5ce99898bddfda64e1769f3aeR196), which does not solves @jmekkatt's observation. Alternatively, when I tested with the `canEdit` property, that solves the problem. However, based on your previous [comment](https://github.com/RedHatInsights/uhc-portal/pull/410#discussion_r3062798250) that's not the recommended property to use. 

I'm conflicted here. Should I use the `canEdit` property, which solves the expectation for @jmekkatt?
Or, should we keep the current property and delegate this fix to a separate effort?

Meanwhile, I created a ticket for the backend to look into the role issue with the channel field [OCM-24237](https://redhat.atlassian.net/browse/OCM-24237)

| (Current) `canUpdateClusterResource` | `canEdit` |
| ----------------------------- | -------------------- |
| Screenshot 2026-04-28 at 6 27 17 PM | Screenshot 2026-04-28 at 5 28 27 PM |


[OCM-24237]: https://redhat.atlassian.net/browse/OCM-24237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jmekkatt:
Hi @t-hendricks , Regarding your comment
>The TLDR; we're trying to avoid adding any filters within client side code and just display the returned data from the API. The backend also has the ability to gate unstable channels. Rather than testing with a feature flag on the client-side, if possible, we should try testing these features with the capabilities removed. If you have more questions, dm me in slack, so we can avoid large threads.

If we are proceeding with these changes, I would suggest updating the acceptance criteria with clearer expectations.

Additionally, are we planning to remove the ocmui-unstable-cluster-versions feature flag–related logic as part of this PR? If not, it could lead to confusion, since this flag directly influences version behavior.
For example:

- If ocmui-unstable-cluster-versions = true and capability.organization.non_stable_channel_group = true, both version and channel fields are enabled with unstable definitions.
- If ocmui-unstable-cluster-versions = false and capability.organization.non_stable_channel_group = true, the version field does not filter unstable versions, while the channel field still does.

This inconsistency could be confusing. In my opinion, aligning this behavior is important for both Day 1 and Day 2 of this feature.

lizagilman:
Verified with backend team -
1. `canUpdateClusterResource` is the correct one for patching a cluster.  `canEdit` disables the pencil but it doesn't check the correct permission. If adding a machine pool role makes `canUpdateClusterResource` resolve to `true`, this is a backend issue with this role.  
2. Delete protection issue is a ui bug, we are not checking the correct permission. We need to check `update` action on `DeleteProtectionResource` , same as backend does.

jmekkatt:
Hi @t-hendricks,

Another observation: even when the capability.organization.non_stable_channel_group capability is disabled, the "fast" channel is still shown based on the channel definition. However, the UI does not yet support the fast channel. This is being tracked in https://redhat.atlassian.net/browse/OCMUI-4518, which is currently in progress.
I’m not sure if this has any implications on the UI . could you please review?

https://github.com/user-attachments/assets/f1333ba4-8bef-4b58-848e-ed554420c4b1


Also, I think if we remove the _ocmui-unstable-cluster-versions_ feature flag and instead rely solely on the _capability.organization.non_stable_channel_group_ capability, this issue might get resolved as well.

jmekkatt:
> @t-hendricks thanks for the changes.
> 
> Consider the below scenario where the user is trying to perform a cluster upgrade. As per the information available on the epic,
> 
> > It will not be possible to upgrade from Y to Y+1 without changing channel first (from stable-4.15 to stable-4.16 for example). Suggested upgrades and the upgrade process stay unchanged.
> > Given that in order to do Y upgrades it will be necessary to switch channels first, we want to provide a way to change channels in the “Cluster > Settings” tab too. We will add another card before the “Update status” one with the channel switching ability, similarly to what we do in the “Cluster > Overview” tab.
> 
> I created a HCP cluster in EUS channel with version 4.16.34 keeping the other options default.
> 
> The current behavior of the cluster upgrade process presents some issues: **Channel Switching Failure**: After the cluster was ready, an attempt to switch the upgrade channel to stable/fast resulted in a backend error. **Incorrect User Experience**: Users are shown a list of channel options but are prevented from selecting a desired version. This is confusing, as the system should allow channel changes and then display the appropriate list of available upgrades for that new channel. **Version Availability Discrepancy**: When attempting to switch to the eus-4.18 channel, the upgrade list correctly showed 4.17 versions as available. However, the system prevented the user from actually switching to the stable 4.17 channel.
> 
> Image Image
> Kindly investigate this behavior. Attached my findings here.
> 
>  Screen.Recording.2026-04-20.at.12.26.03.PM.mov 
>  Screen.Recording.2026-04-20.at.12.17.11.PM.mov

Hi @t-hendricks , In addition to @shivanthi-amara’s observation, I think we need some clarity on the following behavior as well:
For example, a user can create a ROSA hosted cluster with version 4.16.36 (EUS) and channel stable-4.18 on day 1. However, for another ROSA hosted cluster created with version 4.16.36 (EUS) and channel eus-4.16, the user is unable to modify the channel to stable-4.18 during a day 2 operation.
Could you please clarify if this behavior is expected? Please refer to the attached recording for more details.

https://github.com/user-attachments/assets/f48b270f-b2d8-4e5d-ad35-4733457bddf5

t-hendricks:
> Hi @t-hendricks,
> 
> Another observation: even when the capability.organization.non_stable_channel_group capability is disabled, the "fast" channel is still shown based on the channel definition. However, the UI does not yet support the fast channel. This is being tracked in https://redhat.atlassian.net/browse/OCMUI-4518, which is currently in progress. I’m not sure if this has any implications on the UI . could you please review?
> 
> Also, I think if we remove the _ocmui-unstable-cluster-versions_ feature flag and instead rely solely on the _capability.organization.non_stable_channel_group_ capability, this issue might get resolved as well.

Hey @jmekkatt, I received confirmation from CS that the `fast` channel is enabled in the lower environments, such as int and stage. Therefore, it is no longer hidden behind the non_stable capability. The other channels (e.g. `candidate` and `nightly`) should be hidden when the capability is disabled.

t-hendricks:
> @t-hendricks thanks for the changes.
> 
> Consider the below scenario where the user is trying to perform a cluster upgrade. As per the information available on the epic,
> 
> > It will not be possible to upgrade from Y to Y+1 without changing channel first (from stable-4.15 to stable-4.16 for example). Suggested upgrades and the upgrade process stay unchanged.
> > Given that in order to do Y upgrades it will be necessary to switch channels first, we want to provide a way to change channels in the “Cluster > Settings” tab too. We will add another card before the “Update status” one with the channel switching ability, similarly to what we do in the “Cluster > Overview” tab.
> 
> I created a HCP cluster in EUS channel with version 4.16.34 keeping the other options default.
> 
> The current behavior of the cluster upgrade process presents some issues: **Channel Switching Failure**: After the cluster was ready, an attempt to switch the upgrade channel to stable/fast resulted in a backend error. **Incorrect User Experience**: Users are shown a list of channel options but are prevented from selecting a desired version. This is confusing, as the system should allow channel changes and then display the appropriate list of available upgrades for that new channel. **Version Availability Discrepancy**: When attempting to switch to the eus-4.18 channel, the upgrade list correctly showed 4.17 versions as available. However, the system prevented the user from actually switching to the stable 4.17 channel.
> 
> ....

@jmekkatt thanks for bringing this up. Backend team is currently trying to fix this issue under this ticket: [OCM-24287](https://redhat.atlassian.net/browse/OCM-24287)

The original solution to bypass the backend error message was to add a new capability within the org: `allow_switch_to_eol_channel_group`

[OCM-24287]: https://redhat.atlassian.net/browse/OCM-24287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

shivanthi-amara:
> Hi @t-hendricks , Updating the channel via "Edit channel" is always blocked with an "unsupported upgrade policy" error when the cluster is set to Recurring updates. If changing the channel isn't supported in this mode, we should disable the edit option and show an appropriate tooltip instead of surfacing a backend error Review the recording
> 
>  Screen.Recording.2026-04-28.at.12.58.59.PM.mov

@jmekkatt Thanks for the scenarios and insightful comments.

@t-hendricks Thank you for correcting this. I understand this scenario was deferred previously due to concerns about the implementation logic.

shivanthi-amara:
@t-hendricks Tested [OCMUI-4282](https://redhat.atlassian.net/browse/OCMUI-4282) for Installable versions API actually carries available_channels. ROSA, OSD cluster creation wizard — Channel field still works and Regression: cluster create still completable. There seem to be no issues with the tests. The testcase is already attached in the jira ticket( although I am having a bit of trouble uploading the screenshots and there exist some definitions as per the previous implementation of EUS channels). Kindly review the test scenarios highlighted in yellow and add a tc-approved label if it looks good to you.



[OCMUI-4282]: https://redhat.atlassian.net/browse/OCMUI-4282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

t-hendricks:
> @t-hendricks Tested [OCMUI-4282](https://redhat.atlassian.net/browse/OCMUI-4282) for Installable versions API actually carries available_channels. ROSA, OSD cluster creation wizard — Channel field still works and Regression: cluster create still completable. There seem to be no issues with the tests. The testcase is already attached in the jira ticket( although I am having a bit of trouble uploading the screenshots and there exist some definitions as per the previous implementation of EUS channels). Kindly review the test scenarios highlighted in yellow and add a tc-approved label if it looks good to you.

Hey @shivanthi-amara, I added the `tc-approved` label to the ticket.

[OCMUI-4282]: https://redhat.atlassian.net/browse/OCMUI-4282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/ChannelEdit/ChannelEdit.tsx
kelvah: https://redhat.atlassian.net/browse/OCMUI-4528

shivanthi-amara (approved):
LGTM

PR #415, opened 2026-03-26
OCMUI-3829: fix horizontal container-overflows

eliranmal:
# Summary

fixes container overflows that cause undesired horizontal-scrollbars, in cluster-details page and inside the update-cluster modal.

also refactors update-cluster modal to simplify impl' and align with the design-system's conventions.

# Jira

resolves [OCMUI-3829][1]

# Additional information

### changes

- remove the `top-row` class from cluster-details container wrapper, and the matching CSS rule from app stylesheet.  
  this rule was introduced a long time ago, as part of the initial cluster-details impl', and apparently served as layout hack in early patternfly versions (see [here][2], [here][3], and [here][4])
- remove a style hack in the in cluster-upgrade modal header - the `width: 100%` rule that was put in place to make the background cover the available width - alongside with the background color and bottom border (that were also part of that hack)

##### off-topic

- remove `ocm-upgrade-wizard__modal` classname and associated CSS rule as extraneous; it doesn't apply, as modal wrapper styles cannot be applied alongside `openshift` classname
- refactor to standardize upgrade-modal header:
  - remove `ModalWizardHeader` and replace it with the patternfly standard `WizardHeader` 
  - add an `onClose` handler in the wizard header, and conditionally declare  `onClose` in the modal, for cases when the wizard header isn't available (i.e. when the wizard is done and 'finished-step' is shown in the modal)
 
_\* initially removed from the modal and only applied to the wizard, to avoid layout issues (the modal-level handler renders a 'close' button which pushes all content to the left).  updated after feedback from QE._

### implications

the cluster-upgrade modal no longer has a header when the final screen is shown.  i.e. after the wizard was submitted and the modal shows an empty-view with the submit status.

see the 'upgrade-modal final screen' screenshots for illustration.

# How to Test

### cluster details

- navigate to cluster-list page
- choose any cluster and navigate to its cluster-details page
- scroll to the bottom and examine the page to verify there's no horizontal scrollbar
- try to scroll horizontally and verify there's no movement, and that no horizontal-scrollbar appears

repeat these checks with different layout breakpoints - narrow, medium, wide.  to cross breakpoints, you can either zoom in/out or resize your browser.  
for reference - the narrow one will show inner elements stacked vertically (like the resource-usage gauges in the 'overview' tab), the wide one will make the app-navigation sidebar always visible.

### upgrade cluster modal

- choose an upgradeable cluster (for example [upgrade-rosa-1][5]) and navigate to its cluster-details page
- navigate to the cluster-details 'settings' tab
- scroll to the bottom and click the 'update' button at the bottom of the 'update status' card
- examine the modal to verify there's no horizontal scrollbar, try to scroll horizontally and verify no scroll occurs 

# Screen Captures

### cluster details scrollbar

before / after:

https://github.com/user-attachments/assets/8ea859d0-ad98-4064-9dee-b6e7d6f60ee0

### upgrade cluster modal scrollbar

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/794348e5-7f83-477f-9853-d0839bf34987" /> |  |

### upgrade-modal final screen

| State | Before | After |
| --- | --- | --- |
| success |  |  |
| pending |  |  |
| error |  |  |

# Review process

## QE Reviewer

[1]: https://issues.redhat.com/browse/OCMUI-3829
[2]: https://github.com/RedHatInsights/uhc-portal/commit/1c3e337dd5ec6c6a2a590657e793ceba377db1a7#diff-1fd86da69dc309c9635372717d250de8288ba6f7a2bfbaa86a01c6a0b1592733L318
[3]: https://github.com/RedHatInsights/uhc-portal/commit/9ec549d7391d0e49d4f06aa9c1805ab041cdf719#diff-1640a9348fef9263e6c088ff406325a17cc80020cb929b5494ebe0ba696e4f4aR224
[4]:https://github.com/RedHatInsights/uhc-portal/commit/42d4b922737a15c9d1af73d13fe23630379517e6#diff-1fd86da69dc309c9635372717d250de8288ba6f7a2bfbaa86a01c6a0b1592733L297
[5]: https://prod.foo.redhat.com:1337/openshift/details/s/3BRfogktfrVUsYnwGM2J9JZe2F7

[OCMUI-3829]: https://redhat.atlassian.net/browse/OCMUI-3829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kelvah (approved):

jmekkatt (changes requested):
Hi @eliranmal , I have a concern regarding the upgrade modal – specifically the final screen behavior.
In cases of backend or unexpected errors, the modal displays an error but does not provide any option for the user to close it. The only way out is to refresh the page/press escape button, which isn’t a good user experience.
We should add a close (x) button at the top right of the modal, similar to what we currently have in staging.
Could you please take a look?

https://github.com/user-attachments/assets/42e7347a-56c0-4e68-86da-429067c4f144

eliranmal:
> Hi @eliranmal , I have a concern regarding the upgrade modal – specifically the final screen behavior. In cases of backend or unexpected errors, the modal displays an error but does not provide any option for the user to close it. The only way out is to refresh the page/press escape button, which isn’t a good user experience. We should add a close (x) button at the top right of the modal, similar to what we currently have in staging. Could you please take a look?

thanx @jmekkatt - good catch.

the 'close' button was removed from modal header as it resulted in regression of the modal layout (as stated in the PR desc' - content gets pushed to the left).

i'll see if it's reasonable to add a 'close' button to the modal body (following the 'success' view convention), otherwise i'll try and tackle that layout problem, or just accept it.

i'll also verify the other states have a 'close' button (i.e. 'pending').

🛠️

eliranmal:
@jmekkatt - i've fixed it by re-adding the modal 'close' button (only when the wizard is done).  i also updated the desc' with screenshots for each 'finished' state.

please take another look, thanx.

meirinetta (approved):
@eliranmal Thanks for the changes, I re-reviewed and LGTM!

jmekkatt (approved):
Verified latest revision and horizontal scrollbars are not seen in cluster details/overview pages, upgrade models.
Additionally tested the upgrade model final screen with different states such as error, pending and success. Changes working as expected. Approving the PR. 
Thank you @eliranmal !

PR #433, opened 2026-04-02
OCMUI-4146: Fix cluster transfer list data when cluster has been deprovisioned

zherman0:
# Summary

If a cluster is deleted while it is in transfer - which should probably not be allowed but alas - we do not show a valid link or information about the cluster because we were pulling information from the cluster api.  The ticket mentions correctly that even deprovisioned will show in the archive page so we do have some data to share. 

That data is contained on the subscription which we can find using the external cluster id.  I therefore changed the fetch for clusterTransferDetails to use the subscription api instead of the cluster api since the subscription seems to stick around indefinitely.  We now can show the cluster display name, the version and the product type if the cluster is deleted.  

Note: External transfer still cannot show this info because it would be restricted from the new owner until after the transfer.

# Jira

Fixes [OCMUI-4146](https://issues.redhat.com/browse/OCMUI-4146)

# How to Test

1. Create a ROSA classic cluster to transfer
```
rosa create cluster --cluster-name test-xfer-1 --region=us-west-2 --sts 
```
2. Transfer the cluster (using the actions drop down, the owner on the details page, or the Access Control tab) to a user in your org (external transfers cannot handle any data like this due to permissions)
3. Check the Cluster Transfer list on the Cluster Requests tab on the Clusters
4. Now delete the cluster
5. Verify the basic info is still there and that there is no longer an actions button (compare to console.dev)

# Screen Captures

| Screenshot From 2026-04-02 08-13-22 | Screenshot From 2026-04-02 08-12-33 |

# Review process

## QE Reviewer

[OCMUI-4146]: https://redhat.atlassian.net/browse/OCMUI-4146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/queries/ClusterDetailsQueries/useFetchSubscriptionsByExternalId.tsx
+++ b/src/queries/ClusterDetailsQueries/useFetchSubscriptionsByExternalId.tsx
@@ -24,7 +24,7 @@ export const useFetchSubscriptionsByExternalId = (externalIds: string) => {
       const response = await accountsService.getSubscriptions({
         filter: subscriptionSearchString,
         page: 1,
-        page_size: 500,
+        page_size: queryConstants.API_PAGE_SIZE,
       });
       return response;
     },
--- a/src/queries/ClusterDetailsQueries/useFetchSubscriptionsByExternalId.tsx
+++ b/src/queries/ClusterDetailsQueries/useFetchSubscriptionsByExternalId.tsx
@@ -1,0 +1,53 @@
+import { useQuery } from '@tanstack/react-query';
+
+import accountsService from '~/services/accountsService';
+
+import { formatErrorData } from '../helpers';
+import { queryConstants } from '../queriesConstants';
+
+/**
+ * Query to fetch subscriptions based on external IDs
+ * @param externalIds comma-separated external IDs wrapped in single quotes (e.g., "'id1','id2','id3'")
+ * @returns query states. Loading, error and subscription data
+ */
+export const useFetchSubscriptionsByExternalId = (externalIds: string) => {
+  const subscriptionSearchString = `external_cluster_id in (${externalIds})`;
+
+  const { isLoading, data, isError, error, isFetching } = useQuery({
+    queryKey: [
+      queryConstants.FETCH_CLUSTER_DETAILS_QUERY_KEY,
+      'accountsService',
+      'subscription',
+      externalIds,
+    ],
+    queryFn: async () => {
+      const response = await accountsService.getSubscriptions({
+        filter: subscriptionSearchString,
+        page: 1,
+        page_size: 500,
+      });
+      return response;
+    },
+    retry: false,
+    enabled: !!externalIds,
+  });
+
+  if (isError) {
+    const formattedError = formatErrorData(isLoading, isError, error);
+    return {
+      data: data?.data,
+      isLoading,
+      isError,
+      error: formattedError,
+      isFetching,
+    };
+  }
+
+  return {
+    data: data?.data,
+    isLoading,
+    isError,
+    error,
+    isFetching,
+  };
+};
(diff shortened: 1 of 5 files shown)


zherman0:
> Hi @zherman0, tested locally, all looks good, except not sure why we are displaying "Deleted" in the Version column? It seems like we could get the version number from the archived details page (it is shown there). Should we update the Status column accordingly "Closed (cluster deleted)" or just "Cluster Deleted" replacing "Closed" in those cases?

@dtaylor113 - Since this is a bug, there is no UX but I tried to clean up the results the best I could. The version is not in the subscription details that I have and the cluster detail no longer exists so I am not sure where the archive page pulls that data.  I could pull in more data perhaps but the overhead does not seem worth it.

As for the Status - that value is programmatically showing the status from the transfer table, I could add some specialized code but again it seems like over kill.  Since the version was not on the subscription and I can see the status of the subscription, I decided to use that to show "Deleted".  

All that said, I am open to suggestions assuming they do not require us to make additional API calls.  The overhead does not seem worth it for an EXTREME edge case of someone deleting a cluster while they are transferring it.

DcooperRH (approved):
Went through and tested and the new field values are working properly

jmekkatt (approved):
Bug fix changes are working as expected. Approved.
Reference recording attached.

https://github.com/user-attachments/assets/8dfb8c67-919a-4011-977f-ca7895c9242f

dtaylor113 (approved):
Tested locally, verified cluster name link works from Cluster Transfer table after cluster has been deleted.  Verified, Version says "Deleted".

Nit. I still would advocate for for something clearer like Version: 'n/a Cluster Deleted', but I agree this might be overkill. 

Code LGTM

src/queries/ClusterDetailsQueries/useFetchSubscriptionsByExternalId.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import { formatErrorData } from '../helpers';
+import { queryConstants } from '../queriesConstants';
+
+/**
+ * Query to fetch subscriptions based on external IDs
+ * @param externalIds comma-separated external IDs wrapped in single quotes (e.g., "'id1','id2','id3'")
+ * @returns query states. Loading, error and subscription data
+ */
+export const useFetchSubscriptionsByExternalId = (externalIds: string) => {
```
kelvah, line 13: I've stumbled on this hook and traced it back to the PR. 

@zherman0  We already have a `useFetchSubscriptionsByClusterId` query. Was it really necessary to duplicate everything just to pass a different filtering criteria?

The filters can be passed as params and applied based on they nature to the api call. We should try to reuse the code we have instead of cloning things to slightly change their behavior.

Also passing the ids as a string with a comma separated list of values is a bad idea.
That's brittle, it's not type safe. The caller has to know the syntax in advance. An array of strings was the way to go. Then we could have mapped them to whatever the APIs are expecting inside the react query hook.

cc the reviewers @dtaylor113 @DcooperRH

zherman0: @kelvah - I understand your concerns with the new hook but there is a  large group of current computer science that purpose built functions are better than handling all kinds of data to handle every case.  I have used both in our code and I worry that people will not understand what they are suppose to send in to do their search.  I definitely would argue this is subjective.  In this case, I liked isolating the functionality as to not have to change the existing function which might have been confusing to other consumers if an existing function changed its definition. Also, the entire file is 44 lines and pure TRQ.

As for passing in an array versus an API purpose built search string -- bah.  I would have had to create an array object at the consumer and then use the exact logic inside the hook to convert the same array I just built back into a string.  Is it more safe?  I could add checks at the caller.  The external IDs are from a different API call so there should not be any invalid data.  The caller has to know the syntax no matter whether it is an array or a comma separated string.  I could really go either way on this but again this is subjective in terms of CS theory.  Thanks.

PR #412, opened 2026-03-23
[NO-JIRA] Update @redhat-cloud-services dependencies

kelvah:
# Summary

Updating all @redhat-cloud-services dependencies as we were lagging behind

Assisted by: Claude/Opus-4.5 

# Jira

NA

# Additional information

While upgrading I realized we were still using a [deprecated](https://github.com/RedHatInsights/frontend-components/blob/7ad7785eaf48df9a0eabf9224b978ab3f1b074f6/packages/components/src/DateFormat/DateFormat.tsx#L23) `DateFormat` from `@redhat-cloud-services/frontend-components`. We had a whole epic to remove them [2871](https://redhat.atlassian.net/browse/OCMUI-2871), not sure why this one was left behind.

I decided to remove it once and for all while upgrading the library to avoid issues.

The component affected is `ClusterHealthStatus`, used only inside `ClusterHealthCard`. It's a card displayed in the monitoring tab of OCP clusters.

The approach I took was to first update the tests for `ClusterHealthCard` to cover all the scenarios, then I replaced `DateFormat` with the default PF component (using one of our wrappers) and then I've run the tests to make sure everything was still fine. 

The only difference from the previous implementation is that for recent dates we were  showing `just now`, now we show `a few seconds ago`. I think we can live with that.

I also added a story for `ClusterHealthCard` because otherwise it's impossible to review all its states.

# How to Test

## ClusterHealthCard

You need to have an OCP cluster to verify `ClusterHealthCard`. You can find one running the app locally and connecting it to the mocked apis:
https://prod.foo.redhat.com:1337/openshift/details/s/1MmOwBoTq1uDmcpBk0NBqJ3dXej#monitoring

You can verify the cluster card by running storybook with `yarn storybook` and then opening its story on the left nav.

## Other changes

There is a major change updating `@redhat-cloud-services/frontend-components` from `6.x` to `7.x`. The only breaking change regards `rbac-client` but we don't use it.

Just spinning the app locally and browsing around should be enough.

The other major update regards types but I don't see any TS issue.

# Screen Captures

The only visible change is in the "Last check-in" value, it is now underlined and it shows the full date on hover. This is the default PF behavior for date-times.

| Before                                              | After                                   |
| --------------------------------------------------- | --------------------------------------- |
| Screenshot 2026-03-23 at 18 30 22 | Screenshot 2026-03-23 at 18 30 40 |

| Story                                              | 
| --------------------------------------------------- |
| Screenshot 2026-04-14 at 12 08 12 |

[...]

PR #438, opened 2026-04-05
[OCMUI-4125]: Migrate yarn to npm

LiorKGOW:
# Summary

This PR migrates the package manager we use, yarn classic, to npm.
In the process, I've removed the yarn-lock file and ran npm install to generate package-lock.json file instead, and we should keep it instead of yarn-lock going forward.

I've posted some followup actions that need to be done after this PR is merged. You are encouraged to read them, and post any suggestions or concerns you might have 😄 

In addition, I've made a comparison between the new package-lock.json file created on this PR and the old yarn.lock file (As of April 6th) - this might change if we have PRs updating packages (package.json & yarn.lock files). 
You can find the comparison file here: https://gist.github.com/LiorKGOW/b540413d737a9d0110880583b2b461bd 
I would recommend viewing the bottom section of the Log in Raw mode to neatly see the output.
It also contains some recommended actions we might want to take after performing the comparison

Assisted by: Claude Sonnet 4.6

# Jira

https://redhat.atlassian.net/browse/OCMUI-4125

# Additional information

As npm is more strict than yarn classic, we get an error when running npm install for the first time due to some dependencies related to assisted installer. As we want to keep things as simple as possible with this migration, I've decided not to address those issues, which need to be handled by the assisted installer team, in this PR. Therefore, I've added the .npmrc file which allows npm to behave like yarn classic does for peer dependencies (making it less strict).

I've opened a [followup ticket](https://redhat.atlassian.net/browse/OCMUI-4374) to track the resolution of the errors presented when running `npm install` without the .npmrc file, which will not be covered in the scope of this PR.

# How to Test

1. Verify the install is clean

Run the following commands: 
`rm -rf node_modules`
`npm install`

Confirms package-lock.json and .npmrc are solid and no peer dep issues sneak in.

2. Check that the tooling scripts work

Linter - `npm run lint`
Formatter - `npm run prettier`

3. Run unit tests

`npm run test`
If all tests pass, the core scripts are working correctly 

4. Run a production build

`npm run build:prod`

5. Run the dev server locally

`npm run start`

6. CI and all other GitHub Workflows are green

# Screen Captures

No visual changes

# Review process

## QE Reviewer


.github/workflows/check-links.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           node-version: "22"
 
       - name: Install dependencies
-        run: yarn install
+        run: npm ci
```
eliranmal, line 22: `ci` is not equivalent to `install`.

please use `npm install` to maintain pairity with the previous impl'.

.github/workflows/check-api.yml
```diff
@@ -43,12 +43,10 @@ jobs:
...
           path: './node_modules'
-          key: check_api-${{ env.NODE_VERSION }}-${{ hashFiles('**/yarn.lock') }}
+          key: check_api-${{ env.NODE_VERSION }}-${{ hashFiles('**/package-lock.json') }}
           
-      - name: Install yarn
-        uses: borales/actions-yarn@v4
-        with:
-          cmd: install --frozen-lockfile --cache-folder .yarn
+      - name: Install dependencies
+        run: npm ci
```
eliranmal, line 49: the yarn command was using a cache (via `--cache-folder`) - is there an equivalent in npm?

if there isn't, we should look deeper into this job's validity (which might rely on there being a package cache, in order to function).

(btw, i think `ci` is probably suitable here, cuz it looks like it serves a similar function to yarn's `--frozen-lockfile`).

.github/workflows/ci.yml
```diff
@@ -28,11 +28,9 @@ jobs:
       - uses: actions/cache@v4
         with:
           path: './node_modules'
-          key: ${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('**/yarn.lock') }}
+          key: ${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('**/package-lock.json') }}
       - name: Install
-        uses: borales/actions-yarn@v4
-        with:
-          cmd: install
+        run: npm ci
```
eliranmal, line 33: same comment as [here][1].


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045664633

.github/workflows/e2e-ci-playwright.yml
```diff
@@ -49,11 +49,9 @@ jobs:
       - uses: actions/cache@v4
         with:
           path: './node_modules'
-          key: ${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('**/yarn.lock') }}
+          key: ${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('**/package-lock.json') }}
       - name: Install
-        uses: borales/actions-yarn@v4
-        with:
-          cmd: install
+        run: npm ci
```
eliranmal, line 54: same comment as [here][1].


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045664633

.github/workflows/e2e-smoke-playwright.yml
```diff
@@ -69,12 +69,10 @@ jobs:
...
         with:
           path: './node_modules'
-          key: ${{ runner.os }}-22.x-${{ hashFiles('**/yarn.lock') }}
+          key: ${{ runner.os }}-22.x-${{ hashFiles('**/package-lock.json') }}
 
       - name: Install dependencies
-        uses: borales/actions-yarn@v4
-        with:
-          cmd: install
+        run: npm ci
```
eliranmal, line 78: same comment as [here][1].


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045664633

.tekton/uhc-portal-pull-request.yaml
```diff
@@ -36,13 +36,12 @@ spec:
     value: |
       #!/bin/bash
       set -ex
-      npm install -g yarn
-      yarn install
+      npm ci
```
eliranmal, line 39: same comment as [here][1].


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045664633

.tekton/uhc-portal-push.yaml
```diff
@@ -34,12 +34,11 @@ spec:
     value: |
       #!/bin/bash
       set -ex
-      npm install -g yarn
-      yarn install --immutable
+      npm ci
```
eliranmal, line 37: same comment as [here][1].

also, what's the npm equivalent of yarn's `--immutable`?

[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045664633

cypress/README.md
```diff
@@ -9,7 +9,7 @@ We run different subsets in different environments, controlled by `{ tags: [...]
 
 These are marked `tags: ['ci']`.
 For running locally, you'll need credentials in environment variables - `CYPRESS_TEST_WITHQUOTA_USER` and `CYPRESS_TEST_WITHQUOTA_PASSWORD` (ask team members), 
-and `yarn start` (or equivalent dev-env) is already running in another terminal.
+and `npm run start` (or equivalent dev-env) is already running in another terminal.
```
eliranmal: no need for `run` in start script

cypress/README.md
```diff
@@ -27,19 +27,19 @@ All these commands require the relevant env vars to be exported in the terminal
...
 ```
-yarn cypress-headless
+npm run cypress-headless
 ```
 
 To execute a specific test in headless mode:
 
 ```
-yarn cypress-headless --spec 'cypress/e2e/RosaClusterWizard.js'
+npm run cypress-headless --spec 'cypress/e2e/RosaClusterWizard.js'
```
eliranmal: this command could probably use an arguments separator (i.e. `--`)

docs/Playwright-e2e-test-automation-faq.md
```diff
@@ -79,31 +79,31 @@ The authentication storage state (`storageState.json`) is created during the glo
 
 ```bash
 # Run a specific test file
-yarn playwright test playwright/e2e/downloads/downloads.spec.ts
+npx playwright test playwright/e2e/downloads/downloads.spec.ts
```
eliranmal: `npx <command>` is not the equivalent of `yarn <command>`.

`npx` runs a package from the registry without the need to install it locally (i.e. installs them globally in a temp location), and executed the package main entrypoint.  
`yarn <command>` is a shortcut to access commands from the package-manager **bin**, which is specific to yarn.

specifically: in our case `playwright` needed to be stated as a dep' in package.json and installed locally, but with `npx` it doesn't/isn't.

please replace `npx` with `npm`, and use the canonical way of running run bin scripts in npm commands.

eliranmal: alternatively (and actually much easier) - replace all these commands with another one, which is available as an npm script (so would have access to bin commnads).

in our case, you can replace `yarn playwright test` with `npm run playwright-headless`.

docs/Playwright-e2e-test-automation-faq.md
```diff
@@ -79,31 +79,31 @@ The authentication storage state (`storageState.json`) is created during the glo
...
-yarn playwright-ui
+npm run playwright-ui
 ```
 
 ## Q: My test passes locally but fails in CI. How do I debug it?
 
 1. **Check the trace:** CI runs generate traces automatically. Download the trace artifact and open it:
    ```bash
-   yarn playwright show-trace test-results/trace.zip
+   npx playwright show-trace test-results/trace.zip
```
eliranmal: same comment as [here][1].


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045838230

docs/Playwright-e2e-test-automation-faq.md
```diff
@@ -79,31 +79,31 @@ The authentication storage state (`storageState.json`) is created during the glo
...
    ```
 2. **Look at screenshots/videos:** Test artifacts (screenshots on failure, videos if configured) are uploaded as CI artifacts.
 3. **Reproduce headless locally:** CI runs headless by default. Test locally in headless mode to rule out headed-vs-headless differences:
    ```bash
-   yarn playwright test playwright/e2e/your-test.spec.ts
+   npx playwright test playwright/e2e/your-test.spec.ts
```
eliranmal: same comment as [here][1].


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045838230

docs/Playwright-e2e-test-automation-faq.md
```diff
@@ -114,21 +114,21 @@ Playwright offers several modes to help you understand what a test is doing step
...
    ```
 
 2. **Debug Mode:** Launches the browser with Playwright Inspector, letting you step through each action one at a time:
 
    ```bash
-   yarn playwright-debug
+   npm run playwright-debug
    # Or for a specific file:
-   PWDEBUG=1 yarn playwright test playwright/e2e/your-test.spec.ts
+   PWDEBUG=1 npx playwright test playwright/e2e/your-test.spec.ts
```
eliranmal: same comment as [here][1].


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045838230

docs/Playwright-e2e-test-automation-guidelines.md
```diff
@@ -789,16 +789,16 @@ test('critical test', { tag: ['@smoke'] }, async () => {
...
-yarn playwright test --grep="@rosa"
+npx playwright test --grep="@rosa"
 
 # Exclude specific tags
-yarn playwright test --grep-invert="@cluster-creation"
+npx playwright test --grep-invert="@cluster-creation"
 
 # Combine tags
-yarn playwright test --grep="@smoke" --grep="@rosa"
+npx playwright test --grep="@smoke" --grep="@rosa"
```
eliranmal: same comment as [here][1].


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045838230

docs/Playwright-e2e-test-automation-faq.md
```diff
@@ -143,8 +143,8 @@ Playwright offers several modes to help you understand what a test is doing step
 
 5. **Trace Viewer:** Run with tracing enabled to capture a detailed timeline of every action, network request, and DOM snapshot:
    ```bash
-   yarn playwright test --trace on
-   yarn playwright show-trace test-results/trace.zip
+   npx playwright test --trace on
+   npx playwright show-trace test-results/trace.zip
```
eliranmal: same comment as [here][1].


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045838230

docs/Playwright-e2e-test-automation-guidelines.md
```diff
@@ -1001,16 +1001,16 @@ test.skip('broken test - JIRA-1234', async () => { ... });
...
-yarn playwright-headed
+npm run playwright-headed
 
 # Debug Mode (Step through)
-yarn playwright-debug
+npm run playwright-debug
 
 # Specific test file
-yarn playwright test playwright/e2e/downloads/downloads.spec.ts --headed
+npx playwright test playwright/e2e/downloads/downloads.spec.ts --headed
```
eliranmal: same comment as [here][1].


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045838230

docs/Playwright-e2e-test-automation-guidelines.md
```diff
@@ -1049,10 +1049,10 @@ test('validate state', async ({ page }) => {
 
 ```bash
 # Run with tracing
-yarn playwright test --trace on
+npx playwright test --trace on
 
 # Open trace viewer
-yarn playwright show-trace test-results/trace.zip
+npx playwright show-trace test-results/trace.zip
```
eliranmal: same comment as [here][1].


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045838230

playwright/README.md
eliranmal, line 1: same comment as [here][1].


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3045838230

.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
eliranmal, line 1: why is this needed?

cypress-qe-triggers-fedramp.sh
```diff
@@ -15,6 +15,6 @@ echo "$TAGS"
 
 # Install all the required plugins/packages for QE execution enviroument.
 # Installation will be done on the dynamic nodes or your machine as per place of call of this file.
-yarn install
+npm ci
```
eliranmal, line 18: `ci` is not equivalent to `install`

cypress-qe-triggers.sh
```diff
@@ -16,7 +16,7 @@ echo "$TAGS"
 
 # Install all the required plugins/packages for QE execution enviroument.
 # Installation will be done on the dynamic nodes or your machine as per place of call of this file.
-yarn install
+npm ci
```
eliranmal, line 19: `ci` is not equivalent to `install`

Makefile
```diff
@@ -20,11 +20,11 @@ setup: node_modules insights-proxy-check
 
 .PHONY: node_modules
 node_modules:
-	yarn install
+	npm ci
```
eliranmal, line 23: `ci` is not equivalent to `install`

package.json
```diff
@@ -192,33 +192,36 @@
...
     "webpack-bundle-analyzer": "^4.10.1"
   },
-  "resolutions": {
+  "overrides": {
     "@cypress/webpack-preprocessor": "6.0.2",
-    "@redhat-cloud-services/**/axios": "^1.12.2",
+    "@redhat-cloud-services/frontend-components-config": { "axios": "^1.12.2" },
+    "@redhat-cloud-services/frontend-components-utilities": { "axios": "^1.12.2" },
+    "@redhat-cloud-services/javascript-clients-shared": { "axios": "^1.12.2" },
+    "@redhat-cloud-services/rbac-client": { "axios": "^1.12.2" },
```
eliranmal: 1. npm allows dynamic declaration of the override version, such that it matches the declared dependency.
  this means we can state `"axios": "$axios"` (because we want to override to the same version that **we** declare).
1. stating individual packages inside `@redhat-cloud-services` is rigid, and brittle (i.e. requires maintenance and not forward compatible).
  is there an option to use glob patterns in npm overrides?  or another syntax that allows to capture a package on any depth?

package.json
```diff
@@ -248,6 +251,6 @@
   },
   "engines": {
     "node": ">=18.12.0 <19.0.0 || >=20.0.0 <21.0.0 || >=22.0.0",
-    "yarn": ">= 1.22.19"
+    "npm": ">= 10"
```
eliranmal: we shouldn't state a restriction for npm version - npm versions are bound to the node version used.

(also this restriction might break things as our node version range is quite wide.  e.g. if node 18 doesn't arrive with npm 10)

README.md
```diff
@@ -19,12 +19,12 @@ Slack channels: `#forum-cluster-management` for OCM in general, `#ocm-osd-ui` fo
 ## Requirements
 
 - [NodeJS](https://nodejs.org/) `>=18.12.0 <19.0.0 || >=20.0.0 <21.0.0 || >=22.0.0`
-- [Yarn](https://classic.yarnpkg.com/lang/en/) `1.22.19` - (higher versions are not supported)
+- [npm](https://www.npmjs.com/) `>= 10` (comes with Node.js)
```
eliranmal: no need to state npm version if node version is stated.

README.md
```diff
@@ -19,12 +19,12 @@ Slack channels: `#forum-cluster-management` for OCM in general, `#ocm-osd-ui` fo
...
 - [Podman](https://podman.io/docs/installation) `>= v5.5.2`
 
 ## Setup
 
-For a first time setup, it's required to run `yarn fec patch-etc-hosts`.
+For a first time setup, it's required to run `npx fec patch-etc-hosts`.
```
eliranmal: `npx` is not equivalent to `yarn`

README.md
```diff
@@ -49,18 +49,18 @@ yarn build
...
 
 > **Note:** This script requires Red Hat VPN connection to access the UI.
 
 ```bash
-yarn install && yarn dev
+npm install && npm run dev
 ```
 
-### `yarn start`
+### `npm run start`
```
eliranmal: no need for `run` in start script

eliranmal (changes requested):
✏️

.tekton/uhc-portal-push.yaml
LiorKGOW: The closest npm equivalent to yarn install --immutable (or yarn install --immutable) is the npm ci command: 
https://github.com/npm/rfcs/discussions/388#discussioncomment-833788

cypress/README.md
LiorKGOW: Suggestion added
Thanks Lyn !

.npmrc
LiorKGOW: This allows npm to behave like yarn classic did. 
It treats peer dependencies as optional warnings rather than hard requirements, meaning to ignore peer dependency conflicts during installation rather than failing.

As I had some errors when I ran npm install, we decided to not address them on this PR (some relate to assisted installer needing to update their deps) - we as Roberto and I. 
Although It is better to address those, we are going to be blocked on assisted installer updating their packages (as we asked them already several times, but are delaying it). As this is needed by the end of Q1, we are going to need to address them later on. 

With this PR, this ensures the functionality of the newly used npm to be equivalent to the previous yarn classic we were using previously.

package.json
LiorKGOW: So, just to make sure I understand your proposal correctly, are you suggesting to drop this requirement all together? 
```suggestion
```

README.md
LiorKGOW: So, just to make sure I understand your proposal correctly, are you suggesting to drop this line all together?
```suggestion
```

README.md
LiorKGOW: Switched to npm

.github/workflows/check-api.yml
LiorKGOW: `npm ci` already implies frozen lockfile behavior by design. The cache folder flag was a yarn-specific concern and doesn't need to be carried over.

.tekton/uhc-portal-push.yaml
LiorKGOW: --immutable is the Yarn Berry (v2+) equivalent of --frozen-lockfile. Both mean "fail if the lockfile would need to be updated". In yarn Classic, what we were using, --immutable is not a valid flag, it doesn't exist. The Yarn classic equivalent of --frozen-lockfile is npm ci; hence my suggestions here.

docs/Playwright-e2e-test-automation-faq.md
LiorKGOW: Added your second suggestion

.npmrc
eliranmal: got it - thanx for elaborating.

please make sure all of this is tracked with a jira ticket.

package.json
eliranmal: yeah, exactly ☝️ 👌

README.md
eliranmal: yes, please 👍

.tekton/uhc-portal-pull-request.yaml
```diff
@@ -36,13 +36,12 @@ spec:
     value: |
       #!/bin/bash
       set -ex
-      npm install -g yarn
```
eliranmal, line 39: finally! 👏

removing this line must have felt so good   😁

docs/Playwright-e2e-test-automation-faq.md
```diff
@@ -79,31 +79,31 @@ The authentication storage state (`storageState.json`) is created during the glo
 
 ```bash
 # Run a specific test file
-yarn playwright test playwright/e2e/downloads/downloads.spec.ts
+npm run playwright-headless test playwright/e2e/downloads/downloads.spec.ts
```
eliranmal: please remove `test` here - this command will not run correctly as-is.

docs/Playwright-e2e-test-automation-faq.md
```diff
@@ -79,31 +79,31 @@ The authentication storage state (`storageState.json`) is created during the glo
...
-yarn playwright-ui
+npm run playwright-ui
 ```
 
 ## Q: My test passes locally but fails in CI. How do I debug it?
 
 1. **Check the trace:** CI runs generate traces automatically. Download the trace artifact and open it:
    ```bash
-   yarn playwright show-trace test-results/trace.zip
+   npm exec playwright -- show-trace test-results/trace.zip
```
eliranmal: i think the arguments separator (`--`) should come before the package name, and right after `exec`.

see [the docs][1].

(same goes for other occurrences in the PR).


[1]: https://docs.npmjs.com/cli/v9/commands/npm-exec?v=true#npx-vs-npm-exec

docs/Playwright-e2e-test-automation-faq.md
```diff
@@ -79,31 +79,31 @@ The authentication storage state (`storageState.json`) is created during the glo
...
    ```
 2. **Look at screenshots/videos:** Test artifacts (screenshots on failure, videos if configured) are uploaded as CI artifacts.
 3. **Reproduce headless locally:** CI runs headless by default. Test locally in headless mode to rule out headed-vs-headless differences:
    ```bash
-   yarn playwright test playwright/e2e/your-test.spec.ts
+   npm run playwright-headless playwright/e2e/your-test.spec.ts
```
eliranmal: please use a double hyphen (`--`) between the arguments and the command.

(same goes for other occurrences in the PR).

docs/Playwright-e2e-test-automation-faq.md
eliranmal, line 1: looks like some of the commands weren't tested after the update - please verify all commands work the same way as before (throughout all the files in the PR).

README.md
```diff
@@ -19,12 +19,12 @@ Slack channels: `#forum-cluster-management` for OCM in general, `#ocm-osd-ui` fo
...
 
 - [NodeJS](https://nodejs.org/) `>=18.12.0 <19.0.0 || >=20.0.0 <21.0.0 || >=22.0.0`
-- [Yarn](https://classic.yarnpkg.com/lang/en/) `1.22.19` - (higher versions are not supported)
+- [npm](https://www.npmjs.com/) `>= 10` (comes with Node.js)
 - [Podman](https://podman.io/docs/installation) `>= v5.5.2`
 
 ## Setup
 
-For a first time setup, it's required to run `yarn fec patch-etc-hosts`.
+For a first time setup, it's required to run `npm fec patch-etc-hosts`.
```
eliranmal: please add `run` after `npm`.

.github/workflows/e2e-ci-playwright.yml
```diff
@@ -122,9 +120,9 @@ jobs:
           WORKERS: ${{ github.event.inputs.workers || '4' }}
           BASE_URL: https://prod.foo.redhat.com:1337/openshift/
         run: |
-          LOGGING=quiet yarn start 2>&1 | grep --line-buffered -Ev '\[SERVER\] \[SERVE\]' &
-          NODE_TLS_REJECT_UNAUTHORIZED=0 yarn wait-on -i 5000 -t 120s https://prod.foo.redhat.com:1337/apps/openshift/fed-mods.json
-          NODE_TLS_REJECT_UNAUTHORIZED=0 yarn playwright-e2e-ci --workers=$WORKERS
+          LOGGING=quiet npm run start 2>&1 | grep --line-buffered -Ev '\[SERVER\] \[SERVE\]' &
```
eliranmal: no need for `run` in npm `start` script - please remove it.

(same goes for other occurrences in this PR)

eliranmal (changes requested):
re-reviewed - see inline comments ✏️ 

also, please address open issues from the previous review:
- https://github.com/RedHatInsights/uhc-portal/pull/438#discussion_r3045996286
- https://github.com/RedHatInsights/uhc-portal/pull/438#discussion_r3069675535

(btw, is there any reason the PR is kept in 'draft'?  i think it's safe to move it out of it).

---

> [!important]
>
> i've added some changes to fix broken CI - removed `engines.npm` from package/package-lock file.  make sure to sync your local branch from the remote before pushing additional changes.
>
> actually, just to be on the safe side - after syncing, please re-run `npm install` to see if package-lock updates (if so, push it back to the remote, of course).

package.json
```diff
@@ -248,6 +251,5 @@
   },
   "engines": {
     "node": ">=18.12.0 <19.0.0 || >=20.0.0 <21.0.0 || >=22.0.0",
```
eliranmal:
```suggestion
    "node": ">=18.12.0 <19.0.0 || >=20.0.0 <21.0.0 || >=22.0.0"
```

.tekton/uhc-portal-pull-request.yaml
LiorKGOW: Yes! haha

.npmrc
LiorKGOW: Sure! 
I've opened a [followup ticket](https://redhat.atlassian.net/browse/OCMUI-4374) to track this, and will coordinate with @kelvah and @dtaylor113 @jcaianirh if we already have one. (Mentioned this on the Additional information section in the top comment)

TLDR for the tagged people:
Lyn asked about the purpose of adding the .npmrc file, which has been agreed with @kelvah to make the migration to npm cli easier, making it behave like yarn classic did before this migration (as npm is more strict).
I've opened the mentioned ticket to track the resolution of the npm errors

package.json
LiorKGOW: Looking into this matter, I've read that npm overrides do not support @redhat-cloud-services/* or **/ glob patterns in package name keys. The only wildcard supported is bare "*" to match any package name.
However, your suggestion; a top-level "axios" override, seems to be a good solution.

npm pins axios to our declared version for every package in the dependency tree at any depth, not just the four @redhat-cloud-services/* packages mentioned on the overrides section.
Although it affects **all** packages, not just @redhat-cloud-services/*, as axios is already a direct dependency at ^1.12.2, pinning it globally is more consistent.

(Added as a separate commit)

LiorKGOW:
> re-reviewed - see inline comments ✏️
> ...
> See full comment here: https://github.com/RedHatInsights/uhc-portal/pull/438#pullrequestreview-4095225100

Thanks for the re-review @eliranmal ! 
I've had to log off yesterday, but followed up this morning with answers to the rest of your previous comments and am working today to answer the new ones 🤗 
Thought I should keep the PR is draft while I am making changes in the PR so reviewers won't waste time reviewing code that is going to be changed 😇 
I'll put the PR back to review when all comments have been addressed.

I've fetched the changes you pushed to the remote branch locally. Running `npm install` passes and has not changed package-lock.json. Thanks Lyn !

docs/Playwright-e2e-test-automation-faq.md
LiorKGOW: Thanks for the catch!

docs/Playwright-e2e-test-automation-faq.md
LiorKGOW: Thanks for the catch !

README.md
LiorKGOW: As we don't have a script with this name, the command fails. 
I've decided to change it to `npm exec fec patch-etc-hosts` instead

package.json
eliranmal: > Although it affects all packages, not just @redhat-cloud-services/*

this seems to be diverging from original impl'.  
if that's the case, we shouldn't do it.  we want to maintain pairity with the current impl'.

maybe it's best to use your original proposal (albeit with the `$axios` dynamic ref).

README.md
```diff
@@ -82,8 +81,8 @@ If you see them in the [production source](https://app.segment.com/redhat-devtoo
 By default, the UI runs Assisted Installer without standalone mode. To run with Assisted Installer in standalone mode you need to follow these steps:
 
 - Download https://github.com/openshift-assisted/assisted-installer-app project
-- Inside assisted-installer-app run `npm install && npm run start:federated`
-- In uhc-portal run `LOCAL_APPS=assisted-installer-app:8003 yarn start`
+- Inside assisted-installer-app run `npm install && npm start:federated`
```
eliranmal: as a custom command, `start:federated` can only be called with `run`; please re-add it.

(regardless - this command is not on this repo, so shouldn't be changed).

eliranmal (commented):
✏️   had a new comment, and replied in an open thread [here][1].

another thing: the [node.js upgrade PR][2] is now approved, and only waiting for CI to be fixed upstream.  
once it's merged, you'll need to sync changes and then repeat any manual checks (i.e. verify npm commands work under v24 too).

thanx



[1]: https://github.com/RedHatInsights/uhc-portal/pull/438#discussion_r3074251527
[2]: https://github.com/RedHatInsights/uhc-portal/pull/439

package.json
LiorKGOW: Agreed. Code returned to the original suggestion 👍🏼

README.md
LiorKGOW: Thanks for the catch

LiorKGOW:
> another thing: the [node.js upgrade PR](https://github.com/RedHatInsights/uhc-portal/pull/439) is now approved, and only waiting for CI to be fixed upstream.
> once it's merged, you'll need to sync changes and then repeat any manual checks (i.e. verify npm commands work under v24 too).

Thanks for the heads up! 
I'll work on fetching the latest cahnges and update when the PR is ready for another review

LiorKGOW:
@kelvah @eliranmal @jmekkatt @zherman0 PR is ready for review when you are 😄

.github/workflows/e2e-ci-playwright.yml
```diff
@@ -122,9 +120,9 @@ jobs:
           WORKERS: ${{ github.event.inputs.workers || '4' }}
           BASE_URL: https://prod.foo.redhat.com:1337/openshift/
         run: |
-          LOGGING=quiet yarn start 2>&1 | grep --line-buffered -Ev '\[SERVER\] \[SERVE\]' &
-          NODE_TLS_REJECT_UNAUTHORIZED=0 yarn wait-on -i 5000 -t 120s https://prod.foo.redhat.com:1337/apps/openshift/fed-mods.json
-          NODE_TLS_REJECT_UNAUTHORIZED=0 yarn playwright-e2e-ci --workers=$WORKERS
+          LOGGING=quiet npm start 2>&1 | grep --line-buffered -Ev '\[SERVER\] \[SERVE\]' &
+          NODE_TLS_REJECT_UNAUTHORIZED=0 npm run wait-on -- -i 5000 -t 120s https://prod.foo.redhat.com:1337/apps/openshift/fed-mods.json
```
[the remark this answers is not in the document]
eliranmal: true - we don't have `wait-on` as npm-script.  please use `exec` as suggested ☝️

docs/Playwright-e2e-test-automation-faq.md
eliranmal: it still seems like not all commands were tested (inferred from [this comment][1]).  

please go through all commands (including those in doc files) and execute them to make sure everything works.


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3080540610

eliranmal (changes requested):
see comments ✏️

package.json
eliranmal: what about the use of `$axios` ?

eliranmal (commented):
another comment ✏️

package.json
eliranmal: this somehow got re-added.  please remove `npm` from engines.

eliranmal (commented):
another comment ✏️

cypress/README.md
```diff
@@ -27,19 +27,19 @@ All these commands require the relevant env vars to be exported in the terminal
...
 ```
 
 (Note this shows all tests; you may want to append e.g. `--env grepTags=ci` to see just one subset)
 
 To run Cypress in headless mode:
 
 ```
-yarn cypress-headless
+npm run cypress-headless
 ```
```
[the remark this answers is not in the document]
eliranmal: this is out of scope

.github/workflows/e2e-ci-playwright.yml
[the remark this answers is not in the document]
eliranmal: @LiorKGOW - please have a look at these

package.json
LiorKGOW: Removed

package.json
LiorKGOW: Sorry, I thought you meant discard everything 😅 But I see the point now
I've re-added it

.github/workflows/e2e-ci-playwright.yml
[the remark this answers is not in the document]
LiorKGOW: Suggestions added

.github/workflows/e2e-ci-playwright.yml
```diff
@@ -49,11 +49,9 @@ jobs:
       - uses: actions/cache@v4
         with:
           path: './node_modules'
-          key: ${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('**/yarn.lock') }}
+          key: ${{ runner.os }}-${{ matrix.node-version }}-${{ hashFiles('**/package-lock.json') }}
       - name: Install
-        uses: borales/actions-yarn@v4
-        with:
-          cmd: install
+        run: npm install
```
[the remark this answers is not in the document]
LiorKGOW: I was asked to use npm install to maintain parity with the previous implementation (ci is not equivalent to install).
https://github.com/RedHatInsights/uhc-portal/pull/438#discussion_r3045664633

Let me know if you want me to make an exception for this case @eliranmal

docs/contributing.md
```diff
@@ -94,13 +94,13 @@ To promote consistency in the code base, OCM uses the [Prettier](https://prettie
...
-yarn prettier
+npm run prettier
 ```
 
 To fix all formatting issues in all files in the `src` directory:
 
 ```
-yarn prettier:fix
+npm run prettier:fix
 ```
```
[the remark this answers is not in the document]
LiorKGOW: I want to make the minimum amount of changes to this PR. I'll address only major, non-blocking issues from now on

docs/Playwright-e2e-test-automation-faq.md
LiorKGOW: All commands verified.

LiorKGOW (commented):
Verified all the scripts (also within the documentation files). 

The only thing I wanted to verify, because I haven't run this script before, When running: `npm run sentry:sourcemaps-release` should I provide an `AUTH_TOKEN` ? 
i get this: 

Image

I can verify this works, but I am not sure what to include as a value with the --auth-token argument. 
Should I check this ? (question to reviewers)

.github/workflows/e2e-ci-playwright.yml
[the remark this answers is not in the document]
eliranmal: @LiorKGOW , we want to maintain pairity within this PR, so shouldn't switch to `npm ci` here.

OTOH we do not want to neglect tech-debt, so i'd say create a new ticket to track that change separately, and put it in the draft sprint.

eliranmal:
> The only thing I wanted to verify, because I haven't run this script before, When running: `npm run sentry:sourcemaps-release` should I provide an `AUTH_TOKEN` ? 
> 
> I can verify this works, but I am not sure what to include as a value with the --auth-token argument. Should I check this ? (question to reviewers)

@LiorKGOW - using actual command values isn't that important here.  what's important is verifying the command runner works the same.  if you get the same results (i.e. _'a value is required for ... but none was supplied'_), that should suffice 👌

package.json
```diff
@@ -193,33 +193,36 @@
...
     "webpack-bundle-analyzer": "^4.10.1"
   },
-  "resolutions": {
+  "overrides": {
     "@cypress/webpack-preprocessor": "6.0.2",
-    "@redhat-cloud-services/**/axios": "^1.12.2",
+    "@redhat-cloud-services/frontend-components-config": { "axios": "$axios" },
+    "@redhat-cloud-services/frontend-components-utilities": { "axios": "$axios" },
+    "@redhat-cloud-services/javascript-clients-shared": { "axios": "$axios" },
+    "@redhat-cloud-services/rbac-client": { "axios": "$axios" },
```
eliranmal, line 201: this whole definition is still too rigid/brittle.

firstly, for some reason our declared direct deps for `@redhat-cloud-services` aren't aligned with the definitions in this `overrides` clause.  i'm not sure how it was decided which packages to override.

that is - in `dependencies` and `devDependencies` we have:

```
@redhat-cloud-services/frontend-components
@redhat-cloud-services/frontend-components-config-utilities
@redhat-cloud-services/frontend-components-notifications
@redhat-cloud-services/frontend-components-config
@redhat-cloud-services/tsc-transform-imports
@redhat-cloud-services/types
```

while in `overrides` we have:

```
@redhat-cloud-services/frontend-components-config
@redhat-cloud-services/frontend-components-utilities
@redhat-cloud-services/javascript-clients-shared
@redhat-cloud-services/rbac-client
```

---

more importantly - because dependencies are a graph and not a tree (at least since NPM v3 AFAIK), it's very hard to cover all the cases by overriding specific packages that are transitive-dependencies (like `@redhat-cloud-services/...` instead of directly `axios`).  
see for example line 6319 in [package-lock][1], which resolves `node_modules/@redhat-cloud-services/frontend-components-utilities` to `"axios": "^0.28.1 || ^1.7.0"` (!).

since AFAICT that whole 'resolutions'/'overrides' thing was introduced as response to a CVE in the first place (see d23cd9f74ae077acc29e67a9210fda2077712bf4), we can't afford regression in that area.

i therefore think there's no other option than to try and use a global axios override (i.e. `"axios": "$axios"`), and verify it doesn't break anything.  
even tho technically it'd be out of scope for a package-manager migration, the risk of introducing such regression is too big.




[1]: https://github.com/LiorKGOW/uhc-portal/blob/ae2ff7c365312c3a2e0518f0a4e5d02d77f98baf/package-lock.json

eliranmal (changes requested):
had one comment, but it's blocking; please see inline ✏️

package.json
eliranmal: ⚠️  see [this comment][1] on #452, which deals with a related issue.

please follow that PR to see if you need to sync yours before merging.


[1]: https://github.com/RedHatInsights/uhc-portal/pull/452#discussion_r3094263557

.github/workflows/e2e-ci-playwright.yml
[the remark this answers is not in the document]
LiorKGOW: On the one hand, I want to suggest adding these changes on this PR, skip the process, as we are making changes to the same files, so it would delay us for changes we want to make eventually. 
On the other hand, I do see the point in finishing with the yarn->npm migration first, and only then consider making additional changes (divide the amount of changes to smaller PRs). This would reduce the amount of both time and effort reviewing and testing in general, and in addition, in a case of a regression, a do-over of all the process. I do support the second approach, as this PR is getting quite large and takes longer then I originally expected it to take.

Keeping the discussion alive in case others disagree with the outcome, but opening a ticket in the meantime to address the suggested changes.

package.json
LiorKGOW: "axios": "$axios" forces every package, including those that declared "axios": "^0.28.1", to resolve to ^1.12.2. Axios 0.x -> 1.x has real breaking changes: error handling behavior changed, stricter JSON parsing, [etc..](https://github.com/axios/axios/blob/v1.x/MIGRATION_GUIDE.md#key-changes-summary). This change could introduce silent runtime regressions in third-party packages that were written against 0.x semantics (like the one you mentioned). A per-package override, while more verbose, at least scopes the risk to packages we've explicitly vetted. The global override trades one kind of brittleness (incomplete yet more verbose coverage) for another (forced major upgrades).

Regarding the overrides selection, it wasn't based on which packages are in our direct deps, it was based on which packages in the resolved graph actually declare an axios dependency. 
`npm ls axios` confirms that every package in the overrides section depends on axios, and every package omitted from overrides (including frontend-components, notifications, tsc-transform-imports, types) either doesn't depend on axios at all, or depends on it only through a package that's already overridden. The selection is deliberate, not arbitrary. 

A global "axios": "$axios" override couples the forced version for all transitive dependencies to whatever the top-level axios version is at any given time. That means when PR #452 lands, ^1.15.0 is automatically propagated globally, including to packages like @openshift-assisted/ui-lib that haven't been tested against it. The per-package approach (the current suggestion) makes that propagation explicit and controlled: you consciously choose which packages get the new version, rather than having it happen as a side effect of a top-level bump. (True also for later upgrades for this package too)

I think we should keep the override section as it is now

eliranmal: the fact is that yarn worked differently - it enforced `axios` version to effectively be 1.15.x everywhere (see [here][1]).

deviating from that would be introducing regression (in security, in this case, which is unacceptable).


[1]: https://github.com/meirinetta/uhc-portal/blob/075ca3871c8085593d759e7f8179ab8472637714/yarn.lock#L5979-L5980

eliranmal: #452 did not use global enforce - it [only overridden][1] `axios` under `@redhat-cloud-services`.   we need to find a way to have the equivalent declaration (and outcome) in npm.

(did you try out my first suggestion - to include all the `@redhat-cloud-services/...` explicit dependencies instead of just some of them?  if not, please do so and see if it changes anything).

overrides are a patch anyway, another option is try to upgrade any packages that bring outdated `axios` versions, to possibly get newer ones.



[1]: https://github.com/RedHatInsights/uhc-portal/pull/452/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R198

kelvah: I've added Netta's changes from #452 locally to this branch and I've installed everything with npm.

Doing an `npm ls axios` shows that only 1 version was downloaded (the latest one). All dependencies are using that.

I don't see what is the issue here.

.github/workflows/e2e-ci-playwright.yml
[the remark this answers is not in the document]
kelvah: I don't have a strong opinion. Feel free to move to `npm ci` in a follow up ticket.

I honestly think it was a low hanging fruit, we could have just fixed it here. I think what went wrong was that we immediately dismissed the idea without having a proper discussion about eventual risks or issues. 

It's ok to control the scope and avoid adding too many things. At the same time if there is an easy win at hand it should be discussed, otherwise we just waste time for unclear reasons.

package.json
LiorKGOW: Then that concludes it. I'll keep changes as they are currently
Thanks for the insight @kelvah

.github/workflows/e2e-ci-playwright.yml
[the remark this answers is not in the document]
LiorKGOW: I totally see your reasoning Roberto.
As this PR hasn't merged yet, I don't mind changing them on this PR and save some time for the team 😄
Added in the last commit. 

After reading about their differences [here](https://dev.to/manthanank/npm-i-vs-npm-ci-understanding-the-difference-2pfk#:~:text=json%20or%20npm%2Dshrinkwrap.,json%20file.&text=Differences:,fast%2C%20and%20reliable%20dependency%20installations.), I went ahead and updated all the `npm install` commands for the workflows to `npm ci` for them to perform a clean install (I don't think it would harm anything, as there shouldn't be any issues to them). In case there is an issue with it, please let me know

package.json
eliranmal: the issues:  this PR's changes aren't aligned with netta's PR, and i think the lock file here shows that older versions are still in use.

please see my OP:

https://github.com/RedHatInsights/uhc-portal/pull/438#discussion_r3107210178

eliranmal: i must be missing something, let's discuss this offline please.

LiorKGOW: The PR has been updated with Netta's changes when I responded with [this comment](https://github.com/RedHatInsights/uhc-portal/pull/438#discussion_r3130869517) 😄

eliranmal: we talked and examined the different options.

since netta's changes were already merged, we didn't have to sync any new content into the branch.

i suggested to try and test my earlier assumption that declaring our direct 'redhat-cloud-services' deps should suffice.

we tried removing 'overrides' definitions for the different 'redhat-cloud-services' (nested) packages one by one, and found that it doesn't change the axios version, as reported by `npm ls axios`.  
even removing **all** 'redhat-cloud-services' overrides didn't make a difference - axios is still reported as 1.15.2. 

@LiorKGOW will do deeper checks to verify this, and we'll circle back.

LiorKGOW: **FYI: The changes discussed in this comment ARE NOT INCLUDED in the PR.** 

Lyn and I had a meeting to discuss this a little. 
We tried removing those overrides one by one (& running npm install right after each one) to see if and how it affects the lock file or the outcome of `npm ls axios`

Our findings: 
- Removing them one by one,
(javascript-clients-shared -> rbac-client -> frontend-components-utilities -> frontend-components-config)
Running `npm install` doesn't change the lock file at all
Nor the output of `npm ls axios` (apart from the word overridden next to the package) -> see screenshot attached.

We think the axios version stays the same (`1.15.2` as we see in the output of the command)

Before the removals: 
image

After the removals: 
image

LiorKGOW:
@eliranmal @kelvah @shivanthi-amara PR is ready for another review when you are 😇

.github/workflows/e2e-ci-playwright.yml
```diff
@@ -122,9 +120,9 @@ jobs:
           WORKERS: ${{ github.event.inputs.workers || '4' }}
           BASE_URL: https://prod.foo.redhat.com:1337/openshift/
         run: |
-          LOGGING=quiet yarn start 2>&1 | grep --line-buffered -Ev '\[SERVER\] \[SERVE\]' &
-          NODE_TLS_REJECT_UNAUTHORIZED=0 yarn wait-on -i 5000 -t 120s https://prod.foo.redhat.com:1337/apps/openshift/fed-mods.json
-          NODE_TLS_REJECT_UNAUTHORIZED=0 yarn playwright-e2e-ci --workers=$WORKERS
+          LOGGING=quiet npm start 2>&1 | grep --line-buffered -Ev '\[SERVER\] \[SERVE\]' &
+          NODE_TLS_REJECT_UNAUTHORIZED=0 npm exec -- wait-on -i 5000 -t 120s https://prod.foo.redhat.com:1337/apps/openshift/fed-mods.json
```
kelvah, line 124: not a request for a change, more like a curiosity, why didn't you use `npx` here? it does the same thing and it's shorter

package.json
```diff
@@ -248,7 +251,6 @@
     ]
   },
   "engines": {
-    "node": ">=22.0.0 || >=24.0.0",
-    "yarn": ">= 1.22.19"
+    "node": ">=22.0.0 || >=24.0.0"
```
kelvah, line 253: Out of scope for this PR but these ranges are weird. The second range is already covered by the first one. Maybe the idea was to allow only 22 or 24 (and not 23, which is EOL), but then this is not enough.

kelvah (approved):
I think this PR is ok. It's a big one, I hope I didn't miss too many things.

I've run most of the scripts locally, I've run the application, everything seems ok. I think it's probably a good idea to run all the E2E tests locally + doing some sanity checks with the UI.

package.json
eliranmal: please leave the resolving of threads to the thread author.  closing threads yourself can be confusing for reviewers.  i'll reopen for now

package.json
eliranmal: my bad - i removed other older versions when upgrading local env' to node 24, and didn't think all the way through about the resulting expression.

it's gonna be fixed within the scope of [OCMUI-4346][1].


[1]: https://redhat.atlassian.net/browse/OCMUI-4346

package.json
eliranmal: @LiorKGOW - why resolve this again?  i asked to keep threads open so they're easy to track.  re-reopening

LiorKGOW: Sorry, I have't closed this again since your last comment
I thought we closed that conversation on our meeting. Excuse me if I was mistaken 🙏🏼 
Do you have anything left to add here ? @eliranmal

package.json
LiorKGOW: Sounds good to me
Waiting for @kelvah 's approval to resolve this thread

package.json
eliranmal: ok, no worries.  it got closed again somehow..  maybe gh is doing that for some reason..?

i don't have anything to add at this point, just wanted to keep it open as reference while i review the changes.

thanx

eliranmal: having reviewed again, i don't see any more changes made since we talked.  what were the results of your checks?

were you able to verify that these statements are redundant and that they don't affect axios version?

eliranmal (commented):
see comments (specifically [this one][1]).

also, now #497 is merged, this PR needs to sync again and adapt.  please take a look at the changes in #497.


[1]: https://github.com/RedHatInsights/uhc-portal/pull/438/changes#r3154322178

package.json
kelvah: Folks, this PR has been dragging for a while now, we have >100 comments and it's a pain to update dependencies in the meantime. Can we try to come to an agreement?

The overrides/resolutions could be redundant _now_ that the lock file has been created from scratch. Likely also the "problematic" dependencies caught up with their updates.

In the worst case scenario, the overrides still do no harm. Actually we do get periodic critical CVEs around axios, and recently more often then before. We may end up having to reintroduce them sooner than later if we want to declare CVE resolved on our side.

shivanthi-amara (approved):
I followed the steps in the How to test section of this PR and they are green. LGTM!!

eliranmal:
> I followed the steps in the How to test section of this PR and they are green. LGTM!!

@shivanthi-amara , notice this PR is still a WIP, and there are still conflicts with the base branch that needs resolution.

after the PR content is stabilized, any tests are gonna have to be repeated.

shivanthi-amara:
@eliranmal what's your definition of WIP? The PR is not in draft mode and Lior requested a review a week ago. I have since initiated my review, taking Roberto's feedback into account, and all checks are currently passing. 
@LiorKGOW, please feel free to request a re-review if you think it is required..

eliranmal:
> @eliranmal what's your definition of WIP? The PR is not in draft mode and Lior requested a review a week ago. I have since initiated my review, taking Roberto's feedback into account, and all checks are currently passing. @LiorKGOW, please feel free to request a re-review if you think it is required..

@shivanthi-amara  yes, it is kinda vague, but by "WIP" i meant - there is still an open thread that will potentially spawn additional changes.

what is clear and visible is the PR still needs to have its conflicts resolved.

eliranmal:
@LiorKGOW - note some recent changes you'd have to sync up with:

- https://github.com/RedHatInsights/uhc-portal/pull/412#discussion_r3170417118
- https://github.com/RedHatInsights/uhc-portal/pull/513#discussion_r3219841630

.github/workflows/e2e-ci-playwright.yml
LiorKGOW: I think for this case specifically, I haven't used `npx`, but `npm run` initially and after Lyn left some related comments, I switched it to `npm exec`

It's been some time now since I last handled this PR, but I don't recall a specific reason for using `npm exec` specifically. 

I don't mind changing it to npx and include it on this PR, but I would rather move forward with this PR as it drags on for a long time now, if that is not a blocker @kelvah 🙏🏼

kelvah: no worries, it wasn't a request for changes. even if it were, at this point it's too late.

I think I got the answer in another PR btw. I think at some point we have finally to catch up with using npx. Resolving.

package.json
LiorKGOW: Following your comment @kelvah, I would recommend merging this PR as it is now, and in case of a regression we'll act accordingly with a an appropriate fix. I've already re-verified after the last rebase and all scripts are working correctly, and performed some sanity checks (including various api calls) to verify we don't have a problem with axios. 
Anyway, we are overriding only specific packages, owned by RedHat, which use an axios version that we are aware of (as it is using the one specified on our package.json file). So I see a little risk in it

How does it sound ? @eliranmal @kelvah

eliranmal: @kelvah, it's not the disagreement that delayed this PR - we already agreed on the route forward in an online meeting weeks ago.   it's that this PR got pushed aside by higher priority work.

eliranmal: @LiorKGOW - the bottom line is we don't wanna have an earlier version of axios.

if you verified that `npm ls` outputs newer versions of axios, i.e. > 1.15.1, than it'll suffice, and this thread can be closed (we can look into cleanups later) 👌

.github/workflows/e2e-ci-playwright.yml
eliranmal: please update this to use `npm`

.github/workflows/e2e-smoke-playwright.yml
eliranmal: please update this to use `npm`

eliranmal (changes requested):
see comments ✏️

eliranmal (approved):

PR #443, opened 2026-04-09
OCMUI-3779: Fix Save button enabled when taint Effect changed without key

dtaylor113:
# Summary

Fixes form validation bug where the Save button in Edit Machine Pool modal was incorrectly enabled when users changed the taint Effect dropdown without defining a taint key-value pair.

Created by: Ambient OCMUI bugfix workflow using Claude Sonnet 4.5

# Jira

Fixes [OCMUI-3779](https://issues.redhat.com/browse/OCMUI-3779)

# Additional information

**Root Cause:**  
The taints validation schema in `useMachinePoolFormik.ts` only validated `key` and `value` fields, but not the `effect` field. When a user changed the Effect dropdown (e.g., from "NoSchedule" to "NoExecute") without providing a key, Formik detected a form change (making it "dirty") but the validation still passed, which enabled the Save button despite the incomplete taint definition.

**Fix:**  
Added validation for the `effect` field that:
- Allows the initial empty taint row (single taint with `key='', value='', effect='NoSchedule'`)
- Requires a non-empty key when the effect is changed from default or when key/value are set
- Follows the same pattern as existing `value` field validation

# How to Test

**Manual Testing (Reproducing the bug scenario):**

1. Open any ROSA/OSD cluster in console.dev.redhat.com
2. Navigate to the **Machine Pools** tab
3. Select a machine pool that **does not have any existing taints** defined
4. Click the kebab menu (⋮) and select **Edit**
5. In the Edit Machine Pool modal, scroll to **"Labels, Taints and AWS tags"** section
6. Under the Taints section, change only the **Effect** dropdown (e.g., from "NoSchedule" to "NoExecute")
   - **Do not** enter a key or value
7. Observe the **Save** button state

**Expected behavior (after fix):**
- ✅ Save button remains **disabled** (form is invalid)
- ✅ Enter a taint Key, Save button becomes enabled

**Before fix:**
- ❌ Save button becomes **enabled** (incorrectly)

**Additional scenarios to test:**
- Add a taint key → Save button becomes enabled ✅
- Remove the taint key → Save button becomes disabled again ✅
- Add complete taint (key + value + effect) → Save button enabled ✅

**Unit Test Verification:**
```bash
yarn test src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/hooks/useMachinePoolFormik.test.ts
```

All tests should pass, including the new regression tests.

# Screen Captures

| CleanShot 2026-04-10 at 13 42 46 | CleanShot 2026-04-10 at 13 43 15 |

# Review process

**Requirements:**
- 2 developer approvals
- 1 QE approval
- CodeRabbit review addressed
- All CI checks passing

[OCMUI-3779]: https://redhat.atlassian.net/browse/OCMUI-3779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { useField, useFormikContext } from 'formik';
+import { useField } from 'formik';
 
 import { SelectOption } from '@patternfly/react-core';
 
@@ -12,22 +12,19 @@ export type TaintEffect = 'NoSchedule' | 'NoExecute' | 'PreferNoSchedule';
 type TaintEffectFieldProps = {
   fieldId: string;
   isDisabled: boolean;
+  setFieldTouched?: (field: string, touched: boolean, shouldValidate?: boolean) => void;
 };
 
-const TaintEffectField = ({ fieldId, isDisabled }: TaintEffectFieldProps) => {
+const TaintEffectField = ({ fieldId, isDisabled, setFieldTouched }: TaintEffectFieldProps) => {
   const [field] = useField<TaintEffect>(fieldId);
-  const { setFieldTouched } = useFormikContext();
   const onChangeEffect = useFormikOnChange(fieldId);
 
-  const onChange = React.useCallback(
-    (value: string) => {
-      onChangeEffect(value);
-      // Touch the sibling key field so validation errors display there
-      const keyFieldId = fieldId.replace(/\.effect$/, '.key');
-      setTimeout(() => setFieldTouched(keyFieldId, true, true));
-    },
-    [onChangeEffect, fieldId, setFieldTouched],
-  );
+  const onChange = (value: string) => {
+    onChangeEffect(value);
+    // Touch the sibling key field so validation errors display there
+    const keyFieldId = fieldId.replace(/\.effect$/, '.key');
+    setFieldTouched?.(keyFieldId, true, true);
+  };
 
   return (
     <SelectField value={field.value} fieldId={fieldId} onSelect={onChange} isDisabled={isDisabled}>
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { useField } from 'formik';
+import { useField, useFormikContext } from 'formik';
 
 import { SelectOption } from '@patternfly/react-core';
 
@@ -11,18 +11,17 @@ export type TaintEffect = 'NoSchedule' | 'NoExecute' | 'PreferNoSchedule';
 
 type TaintEffectFieldProps = {
   fieldId: string;
+  keyFieldId: string;
   isDisabled: boolean;
-  setFieldTouched: (field: string, touched: boolean, shouldValidate?: boolean) => void;
 };
 
-const TaintEffectField = ({ fieldId, isDisabled, setFieldTouched }: TaintEffectFieldProps) => {
+const TaintEffectField = ({ fieldId, keyFieldId, isDisabled }: TaintEffectFieldProps) => {
   const [field] = useField<TaintEffect>(fieldId);
+  const { setFieldTouched } = useFormikContext();
   const onChangeEffect = useFormikOnChange(fieldId);
 
   const onChange = (value: string) => {
     onChangeEffect(value);
-    // Touch the sibling key field so validation errors display there
-    const keyFieldId = fieldId.replace(/\.effect$/, '.key');
     setFieldTouched(keyFieldId, true, true);
   };
 
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
@@ -12,7 +12,7 @@ export type TaintEffect = 'NoSchedule' | 'NoExecute' | 'PreferNoSchedule';
 type TaintEffectFieldProps = {
   fieldId: string;
   isDisabled: boolean;
-  setFieldTouched?: (field: string, touched: boolean, shouldValidate?: boolean) => void;
+  setFieldTouched: (field: string, touched: boolean, shouldValidate?: boolean) => void;
 };
 
 const TaintEffectField = ({ fieldId, isDisabled, setFieldTouched }: TaintEffectFieldProps) => {
@@ -23,7 +23,7 @@ const TaintEffectField = ({ fieldId, isDisabled, setFieldTouched }: TaintEffectF
     onChangeEffect(value);
     // Touch the sibling key field so validation errors display there
     const keyFieldId = fieldId.replace(/\.effect$/, '.key');
-    setFieldTouched?.(keyFieldId, true, true);
+    setFieldTouched(keyFieldId, true, true);
   };
 
   return (
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { useField } from 'formik';
+import { useField, useFormikContext } from 'formik';
 
 import { SelectOption } from '@patternfly/react-core';
 
@@ -16,7 +16,19 @@ type TaintEffectFieldProps = {
 
 const TaintEffectField = ({ fieldId, isDisabled }: TaintEffectFieldProps) => {
   const [field] = useField<TaintEffect>(fieldId);
-  const onChange = useFormikOnChange(fieldId);
+  const { setFieldTouched } = useFormikContext();
+  const onChangeEffect = useFormikOnChange(fieldId);
+
+  const onChange = React.useCallback(
+    (value: string) => {
+      onChangeEffect(value);
+      // Touch the sibling key field so validation errors display there
+      const keyFieldId = fieldId.replace(/\.effect$/, '.key');
+      setTimeout(() => setFieldTouched(keyFieldId, true, true));
+    },
+    [onChangeEffect, fieldId, setFieldTouched],
+  );
+
   return (
     <SelectField value={field.value} fieldId={fieldId} onSelect={onChange} isDisabled={isDisabled}>
       <SelectOption value="NoSchedule">NoSchedule</SelectOption>
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/hooks/useMachinePoolFormik.test.ts
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/hooks/useMachinePoolFormik.test.ts
@@ -542,9 +542,12 @@ describe('useMachinePoolFormik', () => {
           taints: [{ key: '', value: 'test', effect: 'NoSchedule' }],
         };
 
-        // Value field validation should fail when key is empty
-        await expect(validationSchema.validateAt('taints[0].value', values)).rejects.toThrow(
-          'Taint key has to be defined',
+        // Effect validation catches this case and reports error on key field
+        await expect(validationSchema.validateAt('taints[0].effect', values)).rejects.toMatchObject(
+          {
+            message: 'Taint key has to be defined',
+            path: 'taints[0].key',
+          },
         );
       });
 
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/hooks/useMachinePoolFormik.test.ts
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/hooks/useMachinePoolFormik.test.ts
@@ -460,7 +460,7 @@ describe('useMachinePoolFormik', () => {
       });
     });
 
-    describe('Regression: OCMUI-3779 - taints validation', () => {
+    describe('taints validation', () => {
       it('allows initial empty taint row with default effect', async () => {
         const { validationSchema } = renderHook(() =>
           useMachinePoolFormik({
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/hooks/useMachinePoolFormik.test.ts
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/hooks/useMachinePoolFormik.test.ts
@@ -459,5 +459,116 @@ describe('useMachinePoolFormik', () => {
         expect(result).toBe('cr-12345678901234567');
       });
     });
+
+    describe('Regression: OCMUI-3779 - taints validation', () => {
+      it('allows initial empty taint row with default effect', async () => {
+        const { validationSchema } = renderHook(() =>
+          useMachinePoolFormik({
+            cluster: defaultCluster,
+            machinePool: defaultMachinePool,
+            machineTypes: defaultMachineTypes,
+            machinePools: defaultMachinePools,
+          }),
+        ).result.current;
+
+        const values = {
+          ...defaultExpectedInitialValues,
+          taints: [{ key: '', value: '', effect: 'NoSchedule' }],
+        };
+
+        // Initial empty taint row should be valid
+        await expect(validationSchema.validateAt('taints[0].effect', values)).resolves.toBe(
+          'NoSchedule',
+        );
+      });
+
+      it('requires taint key when effect is changed from default', async () => {
+        const { validationSchema } = renderHook(() =>
+          useMachinePoolFormik({
+            cluster: defaultCluster,
+            machinePool: defaultMachinePool,
+            machineTypes: defaultMachineTypes,
+            machinePools: defaultMachinePools,
+          }),
+        ).result.current;
+
+        const values = {
+          ...defaultExpectedInitialValues,
+          taints: [{ key: '', value: '', effect: 'NoExecute' }],
+        };
+
+        // Changing effect without key should fail validation
+        await expect(validationSchema.validateAt('taints[0].effect', values)).rejects.toThrow(
+          'Taint key has to be defined',
+        );
+      });
+
+      it('allows taint with effect when key is provided', async () => {
+        const { validationSchema } = renderHook(() =>
+          useMachinePoolFormik({
+            cluster: defaultCluster,
+            machinePool: defaultMachinePool,
+            machineTypes: defaultMachineTypes,
+            machinePools: defaultMachinePools,
+          }),
+        ).result.current;
+
+        const values = {
+          ...defaultExpectedInitialValues,
+          taints: [{ key: 'app', value: '', effect: 'NoExecute' }],
+        };
+
+        // Taint with key and changed effect should be valid
+        await expect(validationSchema.validateAt('taints[0].effect', values)).resolves.toBe(
+          'NoExecute',
+        );
+      });
+
+      it('requires taint key even when effect is default but value is set', async () => {
+        const { validationSchema } = renderHook(() =>
+          useMachinePoolFormik({
+            cluster: defaultCluster,
+            machinePool: defaultMachinePool,
+            machineTypes: defaultMachineTypes,
+            machinePools: defaultMachinePools,
+          }),
+        ).result.current;
+
+        const values = {
+          ...defaultExpectedInitialValues,
+          taints: [{ key: '', value: 'test', effect: 'NoSchedule' }],
+        };
+
+        // Value field validation should fail when key is empty
+        await expect(validationSchema.validateAt('taints[0].value', values)).rejects.toThrow(
+          'Taint key has to be defined',
+        );
+      });
+
+      it('allows complete taint with key, value, and effect', async () => {
+        const { validationSchema } = renderHook(() =>
+          useMachinePoolFormik({
+            cluster: defaultCluster,
+            machinePool: defaultMachinePool,
+            machineTypes: defaultMachineTypes,
+            machinePools: defaultMachinePools,
+          }),
+        ).result.current;
+
+        const values = {
+          ...defaultExpectedInitialValues,
+          taints: [{ key: 'app', value: 'frontend', effect: 'NoSchedule' }],
+        };
+
+        // Complete taint should be valid
+        await expect(validationSchema.validateAt('taints[0].effect', values)).resolves.toBe(
+          'NoSchedule',
+        );
+        await expect(validationSchema.validateAt('taints[0].key', values)).resolves.toBe('app');
+        await expect(validationSchema.validateAt('taints[0].value', values)).resolves.toBe(
+          'frontend',
+        );
+      });
+    });
   });
 });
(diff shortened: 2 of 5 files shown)


zherman0:
@dtaylor113 - Regarding the test instructions, for my ROSA classic cluster, I could not adjust the taints at all on the default machine pool - it was all greyed out.  I had to create a new machine pool and then I could test it.

dtaylor113:
> @dtaylor113 - Regarding the test instructions, for my ROSA classic cluster, I could not adjust the taints at all on the default machine pool - it was all greyed out. I had to create a new machine pool and then I could test it.

Hi @zherman0 , yes, same for me, but the disabled messages clearly indicated what needed to be done.  I don't know if that's a Rosa thing, is it the same for OSD clusters?

zherman0 (approved):
Tested locally and review code and tests. LGTM.

DcooperRH (approved):
LGTM

jmekkatt:
Hi @dtaylor113 , 
With the recent change, the Save button correctly gets disabled as expected. However, the user only sees a red indicator on the tab along with a disabled Save button, without any clear indication of what needs to be fixed.

Steps

- Open the Edit Machine Pool modal
- Navigate to the Labels and Taints tab
- In the default empty taint row, change the Effect dropdown from NoSchedule to NoExecute

Observed Behavior

- The tab shows a red error indicator
- The Save button is disabled
- No inline validation error message is displayed
- User is not informed what exactly needs to be corrected

Expected Behavior

- When the user changes the Effect dropdown on an empty taint row (without providing a key), an inline validation message should be displayed:
- “Taint key has to be defined”
- This message should appear near the Effect dropdown or Key field, clearly guiding the user on what needs to be fixed, instead of only showing a red tab indicator with no explanation.

Screenshot 2026-04-27 at 1 27 39 PM

jmekkatt (changes requested):
Kindly refer https://github.com/RedHatInsights/uhc-portal/pull/443#issuecomment-4325211480

dtaylor113:
Hi @jmekkatt, nice catch.
I updated so that that error message would be displayed upon changing the taint's Effect dropdown
CleanShot 2026-04-29 at 11 40 32

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
```diff
@@ -16,7 +16,19 @@ type TaintEffectFieldProps = {
...
+
+  const onChange = React.useCallback(
+    (value: string) => {
+      onChangeEffect(value);
+      // Touch the sibling key field so validation errors display there
+      const keyFieldId = fieldId.replace(/\.effect$/, '.key');
+      setTimeout(() => setFieldTouched(keyFieldId, true, true));
+    },
+    [onChangeEffect, fieldId, setFieldTouched],
+  );
```
dtaylor113: I added logic to touch the sibling taint key field whenever the taint 
Effect dropdown changes. Formik only shows errors on fields the user 
has interacted with ("touched"), so when someone changes the taint 
Effect dropdown without clicking on the Key field, the error wasn't 
showing. Now when you change the taint effect dropdown, it marks the taint key 
field as touched too, so the validation error displays right away.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/hooks/useMachinePoolFormik.ts
```diff
@@ -369,7 +369,8 @@ const useMachinePoolFormik = ({
 
                 // Otherwise, require key when effect is set
                 if (!taintKey) {
-                  return new Yup.ValidationError('Taint key has to be defined', value, this.path);
+                  const keyPath = this.path?.replace(/\.effect$/, '.key') ?? this.path;
+                  return new Yup.ValidationError('Taint key has to be defined', value, keyPath);
```
dtaylor113, line 381: I changed the validation error path from taints[i].effect to 
taints[i].key so that the error message actually displays to the 
user. The taint Effect dropdown uses SelectField which doesn't 
render Formik errors, but the Key field uses TextField which does. 
So by attaching the error to the key path instead, the "Taint key 
has to be defined" message now shows up under the Key field.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
```diff
@@ -91,7 +92,13 @@ const EditTaintsSection = ({
...
                         </GridItem>
                         <GridItem span={4}>
-                          <TextField fieldId={valueField} isDisabled={!!taintsDisabledReason} />
+                          <div
+                            onBlur={() =>
+                              setFieldTouched && setTimeout(() => setFieldTouched(keyField, true, true))
+                            }
+                          >
+                            <TextField fieldId={valueField} isDisabled={!!taintsDisabledReason} />
+                          </div>
```
dtaylor113, line 97: This handles when user types in value field and no key has been entered; shows "key required" error msg under the key field.

dtaylor113:
Hi @jmekkatt,
"Taint key has to be defined" error message now shows under the Key field when no key defined and user changes the Effects dropdown or enters text into the value field.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
zherman0: Apologies here Dave, I am very confused by this code. Why don't you just use the validated property. You would need to add the validate to the FieldArray component.  You could probably add some logic there that makes sure that the key is set if value is present.

dtaylor113: Hi @zherman0, let me explain the approach:

The goal: Show "Taint key has to be defined" error under the Key field when:

1. User changes the taint Effect dropdown without entering a key
2. User types in the Value field without entering a key

The challenge: Formik only displays errors on fields that have been "touched" (user interacted with). When the user changes Effect or types in Value, the Key field hasn't been touched, so even though the validation error exists on the Key field, it wouldn't display.

The solution: The `setFieldTouched(keyField, true, true)` calls mark the Key field as touched when Effect or Value changes, so the validation error displays immediately under the Key field.

The validation itself is in the Yup schema (same pattern as labels/awsTags). The `setFieldTouched` is just to trigger the error display.   

Happy to discuss if you see a simpler approach

zherman0: These error validations are tricky.  I see what your code is doing and it is a very unique situation.  I am good with it.

jmekkatt (approved):
Changes LGTM. Thank you.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
```diff
@@ -16,7 +16,19 @@ type TaintEffectFieldProps = {
...
   const [field] = useField<TaintEffect>(fieldId);
-  const onChange = useFormikOnChange(fieldId);
+  const { setFieldTouched } = useFormikContext();
+  const onChangeEffect = useFormikOnChange(fieldId);
+
+  const onChange = React.useCallback(
+    (value: string) => {
+      onChangeEffect(value);
+      // Touch the sibling key field so validation errors display there
+      const keyFieldId = fieldId.replace(/\.effect$/, '.key');
```
lizagilman: The parent component already has it, it can be passed as a prop.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
```diff
@@ -16,7 +16,19 @@ type TaintEffectFieldProps = {
...
+      setTimeout(() => setFieldTouched(keyFieldId, true, true));
```
lizagilman: Are you sure `setTimeout` is needed?

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
```diff
@@ -91,7 +92,14 @@ const EditTaintsSection = ({
                           <TextField fieldId={keyField} isDisabled={!!taintsDisabledReason} />
                         </GridItem>
                         <GridItem span={4}>
-                          <TextField fieldId={valueField} isDisabled={!!taintsDisabledReason} />
+                          <div
+                            onBlur={() =>
+                              setFieldTouched &&
```
lizagilman: Can `setFieldTouched` ever resolve to `false`?

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
```diff
@@ -91,7 +92,14 @@ const EditTaintsSection = ({
...
+                              setTimeout(() => setFieldTouched(keyField, true, true))
```
lizagilman: Are you sure `setTimeout` is needed?

lizagilman (changes requested):

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/hooks/useMachinePoolFormik.test.ts
```diff
@@ -394,5 +394,122 @@ describe('useMachinePoolFormik', () => {
         expect(result).toBe('cr-12345678901234567');
       });
     });
+
+    describe('Regression: OCMUI-3779 - taints validation', () => {
```
lizagilman: nit: It is not common to include JIRA ticket IDs in `describe`

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
```diff
@@ -16,7 +16,19 @@ type TaintEffectFieldProps = {
 
 const TaintEffectField = ({ fieldId, isDisabled }: TaintEffectFieldProps) => {
   const [field] = useField<TaintEffect>(fieldId);
-  const onChange = useFormikOnChange(fieldId);
+  const { setFieldTouched } = useFormikContext();
+  const onChangeEffect = useFormikOnChange(fieldId);
+
+  const onChange = React.useCallback(
```
lizagilman: nit: Are you sure `useCallback` is necessary?

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
dtaylor113: Good point, done. Now passed as a prop from EditTaintsSection.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
dtaylor113: Removed it - tests pass without it. It was copied from useFormikOnChange which uses it as a workaround for a Formik bug, but it's not needed here.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
dtaylor113: It can be undefined in tests that mock useField but not useFormikContext. Added a comment to clarify: // Fallback needed for tests that mock useField but not useFormikContext

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
dtaylor113: Removed it - tests pass without it.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/hooks/useMachinePoolFormik.test.ts
dtaylor113: Done, changed to just "taints validation".

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
dtaylor113: Removed it - using a plain function now since it's not passed to memoized children.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
lizagilman: @dtaylor113 I was referring to the key https://github.com/dtaylor113/uhc-portal/blob/main/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx#L83
Sorry for the misunderstanding

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
```diff
@@ -39,6 +39,8 @@ const EditTaintsSection = ({
   machineTypes,
 }: EditTaintsSectionProps) => {
   const [input] = useField<EditMachinePoolValues['taints']>('taints');
+  // Fallback needed for tests that mock useField but not useFormikContext
+  const { setFieldTouched } = useFormikContext() ?? {};
```
lizagilman: It's a workaround.. the tests should provide proper Formik context instead of silencing the error.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
lizagilman: https://github.com/RedHatInsights/uhc-portal/pull/513/changes#diff-26f976466b093dcacf336d054f7efbacd5cbecd33e1503b69f12279536ca50f5R22 :crossed_fingers: 
Actaully it's only needed when touching the same field, not a sibling/other field. And also we already have `useFormikOnChange` that has this workaround built in.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
```diff
@@ -91,12 +93,15 @@ const EditTaintsSection = ({
                           <TextField fieldId={keyField} isDisabled={!!taintsDisabledReason} />
                         </GridItem>
                         <GridItem span={4}>
-                          <TextField fieldId={valueField} isDisabled={!!taintsDisabledReason} />
+                          <div onBlur={() => setFieldTouched?.(keyField, true, true)}>
```
lizagilman: Both the extra `div` and the `onBlur` are workarounds.. usually we shouldn't validate `onBlur` , the validation will fire **every time** user leaves the field, usually this is not the expected behavior. Can we handle all validations on the schema level?

lizagilman (commented):
@dtaylor113 I see a few unrelated stuff on the diff (package.json, build-tools), please check.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
dtaylor113: That's pre-existing code - taints don't have unique IDs, so index is the only practical option. Happy to address in a separate PR if needed.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
dtaylor113: useFormikOnChange only touches its own field. We need to touch the sibling key field when effect/value changes. Would a new hook be preferred, or is the current approach acceptable?

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
dtaylor113: The validation is at the schema level (Yup). The onBlur isn't running validation - it just marks the key field as touched so Formik displays the existing schema-level error. Without it, the error exists but doesn't show because the key field hasn't been interacted with.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
dtaylor113: Done. Updated the test to properly mock useFormikContext and removed the fallback. Also made setFieldTouched a required prop in TaintEffectField since it's now always provided.

dtaylor113:
Hi @lizagilman, 

> @dtaylor113 I see a few unrelated stuff on the diff (package.json, build-tools), please check.

removed from the PR -thanks

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
lizagilman: `setFieldTouched(keyField, true, true)` The third arg - `true` - for `shouldValidate`  makes the validation re-run everytime on every blur.
As I mentioned before the extra `div` is workaround and the `onBlur` is workaround. Let's try to avoid them..
Anyhow, why do we need this validation at all? The form already shown an error if the key field is empty.

image

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/TaintEffectField.tsx
lizagilman: You can spare the string manipulation in L26 if you pass `KeyField` from here -  https://github.com/dtaylor113/uhc-portal/blob/main/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx#L83 as a prop. 
Also, the `EditTaintsSection` was ok as it was before - no need to pass it as a prop. We usually don't do it.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
dtaylor113: Hi @lizagilman,  I've updated the code per your latest feedback.

I wanted to share an observation: looking back at the review history, all 12 of your comments over the past six weeks have been about code style and implementation patterns (prop passing, removing setTimeout/useCallback, test mocking, etc.) rather than functionality. The actual bug fix - showing the validation error when a user changes the Effect dropdown without entering a key - has been working correctly since the PR was first opened.

I've been happy to address each round of feedback, and I understand the value of maintaining consistent patterns in a shared codebase. But I'm wondering if we should discuss as a team where we draw the line on code refinement vs. shipping working fixes. This PR addresses a real UX issue (OCMUI-3779), and it's been in review for over a month now.

Not looking to skip code review - just want to start a conversation about balancing code quality with delivery. What are your thoughts?

lizagilman (approved):

jmekkatt (changes requested):
No immediate inline error when taint value is entered without a key (single taint row)

Steps to reproduce
1. Open the Edit Machine Pool modal
2. Leave the taint key field empty
3. Type a value in the taint value field (e.g., "test")
4. Leave the effect dropdown as "NoSchedule" (default)
5. Observe: no error is displayed inline

Expected behaviour
An error message ("Taint key has to be defined") should appear immediately while the user is editing, just as it did before this change (previously displayed on the value field).

Actual behaviour
There was no error message appeared on the field .
Recording added : 

https://github.com/user-attachments/assets/49330b13-2b1a-4f39-901c-3b54fc823318

Hi @dtaylor113  , please take a look. Thank you

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/sections/EditTaintsSection.tsx
lizagilman: @dtaylor113 Thanks for bringing this up, we can definitely discuss it with the team maybe in retro meeting.

> The actual bug fix ... has been working correctly since the PR was first opened.

While a working fix is a minimum requirement, it is just the starting point, it doesn’t automatically mean the PR is ready to be merged. If the fix didn't work at all, there would be no point in reviewing it further or leaving other comments.

> all 12 of your comments over the past six weeks have been about code style and implementation patterns (prop passing, removing setTimeout/useCallback, test mocking, etc.)

There is a significant difference between code style and implementation patterns. Code style comments are usually nitpicks (unless they explicitly violate our style guidelines), whereas implementation patterns can be  blockers. For example:
- Avoiding unnecessary workarounds: If a proper, cleaner solution exists, relying on a workaround can be a blocking issue.
- Component architecture: Keeping a component minimal and "dumb" is a best practice that can be a blocking comment if.
- Unnecessary timeouts:  Avoiding `setTimeout` unless absolutely necessary is usually a blocking comment. It introduces potential bugs and requires proper cleanup handling. 

IIRC these examples are in our code guidelines, and we have discussed them as a team previously.

All of this is even more critical given the nature of our codebase. Because it is a legacy system, it already suffers from inconsistent patterns, workarounds, and quick fixes. This makes it difficult to maintain and frequently causes bugs. Adding more complexity on top of this isn't great at all. On the contrary, we should strive to clean up the tech debt rather than add to it. 

Regarding the timeline, I agree that the review process took longer than usual, which is far from ideal and I understand how frustrating that can be. From my side higher-priority tasks delayed the review. That said, lower-priority items generally afford us the time to ensure things are done properly. While it certainly shouldn't have taken this long, the priority of a fix is a valid factor when balancing code refinement vs delivery.

As mentioned before, happy to discuss more about it, also with the team.

dtaylor113: Thanks @lizagilman, it is a balancing act of trade-offs.

dtaylor113:
Hi @jmekkatt, nice catch. Thanks for the thorough testing.

This `onBlur` handler was removed in a previous review round where @lizagilman and I thought it was a workaround, but as it turns out, it's required for this specific use case (user types a value without changing the Effect dropdown). Without it, the key field never gets marked as "touched" and Formik won't display the validation error.

I've restored it:
CleanShot 2026-06-12 at 14 34 38

jmekkatt (approved):
lgtm

PR #445, opened 2026-04-09
OCMUI-4338: Adjust toolbar content so the filter labels are always on a second line

zherman0:
# Summary

When the clusters list page is shown in a very large resolution (or you zoom out), the filter setting chips align on the same line as the input fields which is contrary to pattern fly guidelines.

I have added a new toolbar content to ensure the filter chips are always on the second line no matter the resolution.

# Jira

Fixes [OCMUI-4338](https://issues.redhat.com/browse/OCMUI-4338)

# How to Test

1. Go to the clusters list page
2. Set a filter for "Cluster Type = ROSA"
3. Zoom out (ctrl -) until you have enough resolution to have everything fit on one line (50% should work)
4. Verify the filter chips are on the second line

# Screen Captures

| Screenshot From 2026-04-09 09-45-19 | Screenshot From 2026-04-09 09-59-05 |

# Review process

## QE Reviewer

[OCMUI-4338]: https://redhat.atlassian.net/browse/OCMUI-4338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/clusters/ClusterListMultiRegion/ClusterList.jsx
+++ b/src/components/clusters/ClusterListMultiRegion/ClusterList.jsx
@@ -426,12 +426,13 @@ const ClusterList = ({
                   </ToolbarItem>
                 )}
                 <ClusterListActions showTabbedView={showTabbedView} />
+              </ToolbarContent>
+              <ToolbarContent>
                 <ViewOnlyMyClustersToggle
                   view={CLUSTERS_VIEW}
                   bodyContent="Show only the clusters you previously created, or all clusters in your organization."
                   localStorageKey={ONLY_MY_CLUSTERS_TOGGLE_CLUSTERS_LIST}
                 />
-
                 {isRestrictedEnv() ? null : (
                   <ToolbarItem>
                     <ClusterListFilterChipGroup />


t-hendricks (changes requested):
@zherman0 Should the "View only my clusters" toggle be on the toolbar row?

Image

zherman0:
> @zherman0 Should the "View only my clusters" toggle be on the toolbar row?

@t-hendricks @kelvah - I grappled with that one.  I thought people were accustom to seeing it on the second line, so I kept it there, but I am fine moving it.  Since we don't have a UX person, I will leave it up to Roberto.

t-hendricks:
> > @zherman0 Should the "View only my clusters" toggle be on the toolbar row?
> 
> @t-hendricks @kelvah - I grappled with that one. I thought people were accustom to seeing it on the second line, so I kept it there, but I am fine moving it. Since we don't have a UX person, I will leave it up to Roberto.

Oh, I see. I'm using a bigger resolution. So, I was used to seeing it on the toolbar row. Nevertheless, I don't see it as a blocker. Moving the applied filters to a separate row worked after testing locally. I'll approve once a decision is made on the toggle button.

zherman0:
I did want to follow this up a little more @t-hendricks @kelvah .  I run my screen at 1920x1080.  I have no idea if there is a "right" size, but when I make the "View only my clusters" toggle part of the main toolbar content, I see a wrapped toolbar that looks like this:
image

Therefore, I put the "View only my clusters" toggle with filter chips.  I feel that works better for smaller resolutions, although I do not consider 1920 to be small especially on my laptop screen.

Looking at patternfly's example page, they handle this with adjusting the layout based on screen break points.  I think we should accept my original solution and re-address this when we move the "View cluster archives" into the tabs, which will give us more room to work with.

I am open to suggestions. Thanks.

dtaylor113 (approved):
Tested locally and different viewports, LGTM.
Code LGTM
Approving, assuming other comments raised by others in the PR will be addressed

t-hendricks (approved):

kelvah:
@t-hendricks had a good point. Filters should be on one toolbar row and they should wrap only when there's not enough space.

Let's try solve the original problem only.

jmekkatt (approved):
Tested with multiple screen resolutions and different zoom levels. The filter labels (single or multiple filters) are appeared in second line. 
LGTM.

PR #446, opened 2026-04-10
OCMUI-4354: Fix query key for cluster transfer ownership to prevent refetch loop 

zherman0:
# Summary

Fix query key for cluster transfer ownership to prevent refetch loop 

Basically I set the query key to contain the viewOptions for the pagination recently added.  However, this is not needed when  fetching a single cluster transfer but still cause the key to change which then triggered a useEffect later in the code.

The fix is simply to not change the viewOptions for a single cluster since there is no pagination.

AI did help me track this down and here is the full explanation:

> In useFetchClusterTransfer.tsx the React Query queryKey included the full viewOptions object for CLUSTER_TRANSFER_VIEW, even when the request was getClusterTransferByExternalID (cluster details + modal — no pagination).

> At the same time, a useEffect dispatched onSetTotal whenever data?.data?.total was set. That updates Redux → viewOptions changes → query key changes → React Query refetches → effect runs again → repeat. Cancel/restart transfer makes refetches and responses with total more likely, so the bug can look flaky.

# Jira

Fixes [OCMUI-4354](https://issues.redhat.com/browse/OCMUI-4354)

# How to Test

1. Create a ROSA classic cluster using the default values.
2. Once the cluster is in ready status, go to the access control tab.
5. Click on the transfer ownership tab and, using the button Initiate transfer, submit the details (ask offline if you need another id to transfer to).
7. Click on the Cancel transfer button.
9. Open the transfer ownership modal again and fill in the username, account ID, and organization ID details.
11. Click on the initiate transfer button.
12. For good measure, cancel and re-initiate once more.
13. There should be no error that ocurs

# Review process

## QE Reviewer

[OCMUI-4354]: https://redhat.atlassian.net/browse/OCMUI-4354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
+++ b/src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
@@ -81,7 +81,7 @@ describe('useFetchClusterTransfer', () => {
     };
     getClusterTransferByExternalIDSpy.mockResolvedValueOnce({
       data: listPayload,
-    } as never);
+    });
 
     const { result } = renderHook(() => useFetchClusterTransfer({ clusterExternalID: 'ext-1' }));
 
@@ -98,7 +98,7 @@ describe('useFetchClusterTransfer', () => {
 
   it('searches by transfer ID with pagination options from view state', async () => {
     const searchResponse = { data: { items: [], total: 0 } };
-    searchClusterTransfersSpy.mockResolvedValueOnce(searchResponse as never);
+    searchClusterTransfersSpy.mockResolvedValueOnce(searchResponse);
 
     const { result } = renderHook(() => useFetchClusterTransfer({ transferID: 'tr-99' }));
 
@@ -127,7 +127,7 @@ describe('useFetchClusterTransfer', () => {
         },
       }),
     );
-    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } } as never);
+    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } });
 
     const { result } = renderHook(() => useFetchClusterTransfer({ transferID: 't-sort' }));
 
@@ -154,7 +154,7 @@ describe('useFetchClusterTransfer', () => {
         },
       }),
     );
-    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } } as never);
+    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } });
 
     const { result } = renderHook(() => useFetchClusterTransfer({ transferID: 't-asc' }));
 
@@ -178,7 +178,7 @@ describe('useFetchClusterTransfer', () => {
         },
       }),
     );
-    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } } as never);
+    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } });
 
     const { result } = renderHook(() => useFetchClusterTransfer({ filter: `id='x'` }));
 
@@ -191,7 +191,7 @@ describe('useFetchClusterTransfer', () => {
   });
 
   it('uses custom filter when provided', async () => {
-    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } } as never);
+    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } });
 
     const { result } = renderHook(() => useFetchClusterTransfer({ filter: `owner='org-1'` }));
 
@@ -210,7 +210,7 @@ describe('useFetchClusterTransfer', () => {
   it('dispatches total count for list/search mode when total is present', async () => {
     searchClusterTransfersSpy.mockResolvedValueOnce({
       data: { items: [], total: 42 },
-    } as never);
+    });
 
     const { result } = renderHook(() => useFetchClusterTransfer({ transferID: 't1' }));
 
@@ -244,7 +244,7 @@ describe('useFetchClusterTransfer', () => {
         items: [pendingItem, { id: 'other', cluster_uuid: 'ext-p', status: 'completed' }],
         total: 2,
       },
-    } as never);
+    });
 
     const { result } = renderHook(() =>
       useFetchClusterTransfer({
@@ -268,7 +268,7 @@ describe('useFetchClusterTransfer', () => {
     };
     getClusterTransferByExternalIDSpy.mockResolvedValueOnce({
       data: { items: [acceptedItem], total: 1 },
-    } as never);
+    });
 
     const { result } = renderHook(() =>
       useFetchClusterTransfer({
@@ -288,7 +288,7 @@ describe('useFetchClusterTransfer', () => {
         items: [{ id: 'done', cluster_uuid: 'ext-z', status: 'completed' }],
         total: 1,
       },
-    } as never);
+    });
 
     const { result } = renderHook(() =>
       useFetchClusterTransfer({
@@ -303,7 +303,7 @@ describe('useFetchClusterTransfer', () => {
   });
 
   it('does not dispatch total when search response has no total field', async () => {
-    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [] } } as never);
+    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [] } });
 
     const { result } = renderHook(() => useFetchClusterTransfer({ transferID: 'no-total' }));
 
--- a/src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
+++ b/src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
@@ -1,0 +1,222 @@
+import * as reactRedux from 'react-redux';
+
+import { queryClient } from '~/components/App/queryClient';
+import { queryConstants } from '~/queries/queriesConstants';
+import { viewConstants, viewPaginationConstants } from '~/redux/constants';
+import { useGlobalState } from '~/redux/hooks';
+import { accountsService } from '~/services';
+import { renderHook, waitFor } from '~/testUtils';
+import { ViewOptions } from '~/types/types';
+
+import { refetchFetchClusterTransfer, useFetchClusterTransfer } from './useFetchClusterTransfer';
+
+jest.mock('react-redux', () => ({
+  ...jest.requireActual('react-redux'),
+  useDispatch: jest.fn(),
+}));
+
+jest.mock('~/redux/hooks', () => ({
+  useGlobalState: jest.fn(),
+}));
+
+const useGlobalStateMock = useGlobalState as jest.Mock;
+
+const defaultViewOptions: ViewOptions = {
+  currentPage: 1,
+  pageSize: 20,
+  totalCount: 0,
+  totalPages: 0,
+  filter: '',
+  sorting: {
+    sortIndex: 0,
+    sortField: 'created_at',
+    isAscending: false,
+  },
+  flags: {},
+};
+
+describe('useFetchClusterTransfer', () => {
+  const useDispatchMock = jest.spyOn(reactRedux, 'useDispatch');
+  const mockedDispatch = jest.fn();
+  let getClusterTransferByExternalIDSpy: jest.SpyInstance;
+  let searchClusterTransfersSpy: jest.SpyInstance;
+
+  beforeEach(() => {
+    useDispatchMock.mockReturnValue(mockedDispatch);
+    mockedDispatch.mockClear();
+    useGlobalStateMock.mockImplementation((selector: (s: unknown) => unknown) =>
+      selector({
+        viewOptions: {
+          [viewConstants.CLUSTER_TRANSFER_VIEW]: defaultViewOptions,
+        },
+      }),
+    );
+    getClusterTransferByExternalIDSpy = jest.spyOn(
+      accountsService,
+      'getClusterTransferByExternalID',
+    );
+    searchClusterTransfersSpy = jest.spyOn(accountsService, 'searchClusterTransfers');
+  });
+
+  afterEach(() => {
+    getClusterTransferByExternalIDSpy.mockRestore();
+    searchClusterTransfersSpy.mockRestore();
+  });
+
+  it('does not fetch when transferID, clusterExternalID, and filter are all missing', async () => {
+    const { result } = renderHook(() => useFetchClusterTransfer({}));
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(getClusterTransferByExternalIDSpy).not.toHaveBeenCalled();
+    expect(searchClusterTransfersSpy).not.toHaveBeenCalled();
+  });
+
+  it('fetches by cluster external ID and returns list data', async () => {
+    const listPayload = {
+      items: [{ id: 't1', cluster_uuid: 'ext-1', status: 'completed' }],
+      total: 1,
+    };
+    getClusterTransferByExternalIDSpy.mockResolvedValueOnce({
+      data: listPayload,
+    } as never);
+
+    const { result } = renderHook(() => useFetchClusterTransfer({ clusterExternalID: 'ext-1' }));
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(getClusterTransferByExternalIDSpy).toHaveBeenCalledWith('ext-1');
+    expect(searchClusterTransfersSpy).not.toHaveBeenCalled();
+    expect(result.current.isError).toBe(false);
+    expect(result.current.data).toEqual(listPayload);
+    expect(mockedDispatch).not.toHaveBeenCalled();
+  });
+
+  it('searches by transfer ID with pagination options from view state', async () => {
+    const searchResponse = { data: { items: [], total: 0 } };
+    searchClusterTransfersSpy.mockResolvedValueOnce(searchResponse as never);
+
+    const { result } = renderHook(() => useFetchClusterTransfer({ transferID: 'tr-99' }));
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(searchClusterTransfersSpy).toHaveBeenCalledWith({
+      filter: `id='tr-99'`,
+      page: 1,
+      size: 20,
+      orderBy: 'created_at desc',
+    });
+    expect(getClusterTransferByExternalIDSpy).not.toHaveBeenCalled();
+    expect(result.current.data).toEqual(searchResponse.data);
+  });
+
+  it('uses custom filter when provided', async () => {
+    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } } as never);
+
+    const { result } = renderHook(() => useFetchClusterTransfer({ filter: `owner='org-1'` }));
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(searchClusterTransfersSpy).toHaveBeenCalledWith(
+      expect.objectContaining({
+        filter: `owner='org-1'`,
+      }),
+    );
+    expect(result.current.isError).toBe(false);
+  });
+
+  it('dispatches total count for list/search mode when total is present', async () => {
+    searchClusterTransfersSpy.mockResolvedValueOnce({
+      data: { items: [], total: 42 },
+    } as never);
+
+    const { result } = renderHook(() => useFetchClusterTransfer({ transferID: 't1' }));
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    await waitFor(() => {
+      expect(mockedDispatch).toHaveBeenCalled();
+    });
+
+    expect(mockedDispatch).toHaveBeenCalledWith(
+      expect.objectContaining({
+        type: viewPaginationConstants.SET_TOTAL_ITEMS,
+        payload: expect.objectContaining({
+          viewType: viewConstants.CLUSTER_TRANSFER_VIEW,
+          totalCount: 42,
+        }),
+      }),
+    );
+  });
+
+  it('returns pending transfer slice when showPendingTransfer is true', async () => {
+    const pendingItem = {
+      id: 'tp',
+      cluster_uuid: 'ext-p',
+      status: 'pending',
+    };
+    getClusterTransferByExternalIDSpy.mockResolvedValueOnce({
+      data: {
+        items: [pendingItem, { id: 'other', cluster_uuid: 'ext-p', status: 'completed' }],
+        total: 2,
+      },
+    } as never);
+
+    const { result } = renderHook(() =>
+      useFetchClusterTransfer({
+        clusterExternalID: 'ext-p',
+        showPendingTransfer: true,
+      }),
+    );
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(result.current.data).toEqual({ items: [pendingItem] });
+  });
+
+  it('returns formatted error when the request fails', async () => {
+    getClusterTransferByExternalIDSpy.mockRejectedValueOnce({
+      name: 403,
+      message: 'Forbidden',
+    });
+
+    const { result } = renderHook(() => useFetchClusterTransfer({ clusterExternalID: 'ext-err' }));
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(result.current.isError).toBe(true);
+    expect(result.current.error).toEqual(
+      expect.objectContaining({
+        isError: true,
+        error: expect.objectContaining({
+          message: 'Forbidden',
+        }),
+      }),
+    );
+  });
+});
+
+describe('refetchFetchClusterTransfer', () => {
+  it('invalidates cluster transfer queries', () => {
+    const spy = jest.spyOn(queryClient, 'invalidateQueries');
+    refetchFetchClusterTransfer();
+    expect(spy).toHaveBeenCalledWith({
+      queryKey: [queryConstants.FETCH_CLUSTER_DETAILS_QUERY_KEY, 'fetchClusterTransfer'],
+    });
+    spy.mockRestore();
+  });
+});
--- a/src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
+++ b/src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
@@ -233,7 +233,7 @@ describe('useFetchClusterTransfer', () => {
     );
   });
 
-  it('returns pending transfer slice when showPendingTransfer is true', async () => {
+  it('returns pending transfer items when showPendingTransfer is true', async () => {
     const pendingItem = {
       id: 'tp',
       cluster_uuid: 'ext-p',
@@ -260,7 +260,7 @@ describe('useFetchClusterTransfer', () => {
     expect(result.current.data).toEqual({ items: [pendingItem] });
   });
 
-  it('returns accepted transfer slice when showPendingTransfer matches accepted status', async () => {
+  it('returns accepted transfer items when showPendingTransfer matches accepted status', async () => {
     const acceptedItem = {
       id: 'ta',
       cluster_uuid: 'ext-a',
--- a/src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
+++ b/src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
@@ -116,6 +116,80 @@ describe('useFetchClusterTransfer', () => {
     expect(result.current.data).toEqual(searchResponse.data);
   });
 
+  it('uses updated_at desc when sort field is empty', async () => {
+    useGlobalStateMock.mockImplementation((selector: (s: unknown) => unknown) =>
+      selector({
+        viewOptions: {
+          [viewConstants.CLUSTER_TRANSFER_VIEW]: {
+            ...defaultViewOptions,
+            sorting: { ...defaultViewOptions.sorting, sortField: '' },
+          },
+        },
+      }),
+    );
+    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } } as never);
+
+    const { result } = renderHook(() => useFetchClusterTransfer({ transferID: 't-sort' }));
+
+    await waitFor(() => expect(result.current.isLoading).toBe(false));
+
+    expect(searchClusterTransfersSpy).toHaveBeenCalledWith(
+      expect.objectContaining({ orderBy: 'updated_at desc' }),
+    );
+    expect(result.current.isError).toBe(false);
+  });
+
+  it('uses ascending order when view sort is ascending', async () => {
+    useGlobalStateMock.mockImplementation((selector: (s: unknown) => unknown) =>
+      selector({
+        viewOptions: {
+          [viewConstants.CLUSTER_TRANSFER_VIEW]: {
+            ...defaultViewOptions,
+            sorting: {
+              ...defaultViewOptions.sorting,
+              sortField: 'name',
+              isAscending: true,
+            },
+          },
+        },
+      }),
+    );
+    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } } as never);
+
+    const { result } = renderHook(() => useFetchClusterTransfer({ transferID: 't-asc' }));
+
+    await waitFor(() => expect(result.current.isLoading).toBe(false));
+
+    expect(searchClusterTransfersSpy).toHaveBeenCalledWith(
+      expect.objectContaining({ orderBy: 'name asc' }),
+    );
+    expect(result.current.isError).toBe(false);
+  });
+
+  it('falls back to page 1 and size 20 when view options are zero', async () => {
+    useGlobalStateMock.mockImplementation((selector: (s: unknown) => unknown) =>
+      selector({
+        viewOptions: {
+          [viewConstants.CLUSTER_TRANSFER_VIEW]: {
+            ...defaultViewOptions,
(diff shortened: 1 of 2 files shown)


src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.tsx
```diff
@@ -34,12 +34,17 @@ export const useFetchClusterTransfer = ({
 
   const dispatch = useDispatch();
 
+  /* Paginated search uses viewOptions in the request; external-ID lookup does not. Putting the
+   * whole viewOptions slice in the key for external-ID fetches tied query identity to Redux: this
+   * hook dispatches onSetTotal → viewOptions changes → new queryKey → refetch → loop (React #185). */
```
dtaylor113: Nit. I liked this comment up until the last sentence.  I'm not sure commenting on triggers and "(React #185)" <-- is this OCMUI PR? ... is needed, but I do like the other part of the comment block.

dtaylor113 (approved):
Is it worth adding a unit test for this fix?
Tested thru 3 rounds of initiate -> cancel -> initiate, no errors.
code LGTM.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.tsx
```diff
@@ -61,12 +66,13 @@ export const useFetchClusterTransfer = ({
     refetchInterval: queryConstants.STALE_TIME_60_SEC,
   });
 
-  // Recalculate totalPages when pageSize changes or new data arrives
+  // Recalculate totalPages when pageSize changes or new data arrives (list/search only).
   useEffect(() => {
+    if (clusterExternalID) return;
```
LiorKGOW, line 69: I see you add to this useEffect that reevaluates totalPages when some of the props change. 
As [our code guidelines mention](https://github.com/RedHatInsights/uhc-portal/blob/main/docs/code-guide.md#more-on-useeffect): "...where useEffect is NOT needed: ... Updating state based on props/state changes (calculate during render)"

We should avoid adding more complexity to this useEffect, as it is complex enough for you, the author of this code, for having to add a comment explaining about the flow on line 37-39. 

In addition, shouldn't this change be dependent on filter too? As we might affect the number of resuls and therefore on the totalPages as well

zherman0: @LiorKGOW - Thanks for the insightful comments; I appreciate your thoroughness. As for the useEffect, quoting the guide:
```
Only use useEffect for:

    Synchronizing with external systems (APIs, DOM, third-party libraries)
```
Redux is an external third party library and that is what I am updating here.

As for complexity, this simple addition makes the code more efficient so it is a net positive.  As for "complex because I have comments", LOL.  Comments make things better, not more complex.  Our ancient redux-based pagination makes things more complex.

As for 'filter', at this point the filter has already influence the returned data and I do have `data?.data?.total` which reflects that change so filter would be superfluous.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.tsx
zherman0: Ah yes, sorry @dtaylor113, without context of the original error message for this bug that comment makes no sense.  When I wrote it I was looking at the "Minified React error #185" for the ticket and it meant something to me in the moment.  I will remove it.

zherman0:
@dtaylor113 - Added unit tests

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.tsx
LiorKGOW: I wouldn't say this addition makes the code more efficient, as we add another condition which would get calculated every time this effect runs. 
As for the "happy path" of the condition you added, yes that would be more efficient, but that is only for this edge case.

So, just to make things clear and please correct me if I'm wrong, I see that data is fetched using React Query. Therefore it is handled with ReactQuery. But, we hold the totalPages value in Redux. Why exactly? 
Why do we even need redux for this? This is relevant just for the current component, so shouldn't we hold a local state in the component which needs it (or in this hook) and handle it there? 
I really can't see a reason to include redux here, unless we want to increase the complexity of course

(I know you don't add this code in this PR, I am just trying to initiate a healthy conversation about the code, and see how we can improve things to make a more readable efficient and error prune project 😄)

zherman0: Great stuff here.  First efficiency: we can call this hook with either 1 cluster and get 1 transfer or we can get all cluster transfers by not sending a specific clusterID.  If we are only getting 1 cluster transfer, then we do not need the overhead of calling the dispatch, then calling the redux, then going through the reudx state to set the total count.  So this one line check prevents us going through all that to set the value to 1 and it is not even used. Much more efficient wouldn't you agree?  Additionally, this is not an edge case, ClusterTransferList calls this with no specific cluster (and thus needs pagination).  All the rest of times this gets called with a specific cluster (editing transfer, viewing transfer, etc).  I am not sure the break down in usage, but this is not edge.

Second: Redux and TRQ.  Everything data wise with ClusterTransfers is react-query.  Sadly, the entire pagination  system is still redux.   I would have had to rewrite that entire system and that was not in scope and would have affected every pagination we had in OCMUI.  I will not get into the details here about viewOptions and pagination, but I do encourage you to investigate it and maybe rewrite it as it is needed to finally get rid of redux.

Please keep the questions coming.

LiorKGOW: On efficiency / the guard clause: You're right, I misread the impact. The guard skips the entire dispatch path for single-cluster lookups, not just a cheap condition check. Given the two distinct call patterns (ClusterTransferList vs. individual transfer views), the check is justified. 

On filter: Also fair. data.data.total is already the server-filtered count, so adding filter to the dependency array would be redundant. (Originally I understood that, just asked out of curiosity and general flow of things as filter might affect the number of results and therefore the totalPages as well)

On useEffect + Redux: You clarified my thinking here. The guideline "don't use useEffect to update state based on props/state" targets local useState, calling dispatch during render would be a side effect violation, so useEffect is actually the right call when the target is Redux. However, that just surfaces the real question more sharply: 
totalPages is derived purely from data.data.total, a value already managed by React Query. The only reason a useEffect -> dispatch -> Redux -> re-render round-trip is needed is because the pagination UI reads from Redux. That's the root coupling. Could totalPages be passed down as a prop from this hook instead of read from Redux? That would eliminate the need for an effect entirely.

I understand that's a larger refactor and not in scope here. What would you think about openning a follow-up ticket to decouple the pagination state from Redux for this feature?

zherman0: Feel free to open another ticket, but the decoupling would not be unique to clusterTransfer.  The other big consumers of pagination are the cluster list and access request (and others) and those are also using react query.  

FWIW, I did try hard to not use redux and tried to circumvent this requirement, but the pagination component and code is all intertwined with redux.

LiorKGOW: > I did try hard to not use redux and tried to circumvent this requirement, but the pagination component and code is all intertwined with redux.

Sounds like a great reason to me for a bigger refactor instead of just adding more tech-debt over already existing tech-debt... Refactoring this area of the code, before working on this feature, would have been the better approach IMHO, but that is water under the bridge now. 
It would be great if you could open the ticket and provide some context as you already worked in this area before and know what's going on there 😄 
(Ex. Explain the flow, exactly what needs to be refactored out of redux - including all the relevant areas this is being used, if you considered alternatives)

I suggest to at least open a ticket and raising that to be worked on. 
As we should also tackle tech debt equally as feature work, I would even consider prioritizing this, especially if we want to get rid of redux any time soon 
cc @kelvah ;

LiorKGOW (commented):
In order to test this locally, as reviewers, we need a user which is not ours, including an Account ID and an organization ID. 
What have you used to test this locally @dtaylor113 ? Could you share? 
I would suggest adding it, if possible to the how to test section for other reviewers to test as well as it would ease the review process for everyone 😄

dtaylor113:
Hi @LiorKGOW,
I sent you Zac's Cluster Transfer info. that I used in a slack DM

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.tsx
zherman0: I see the confusion, this is not a feature request, this is a bug fix.  

All the discussion is very valid, but can you please test/review this specific bug as the scope is very specific.  Reach out to me for an ID to transfer to, as stated in the test instructions.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
```diff
@@ -0,0 +1,349 @@
...
+    const { result } = renderHook(() => useFetchClusterTransfer({ clusterExternalID: 'ext-1' }));
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(getClusterTransferByExternalIDSpy).toHaveBeenCalledWith('ext-1');
+    expect(searchClusterTransfersSpy).not.toHaveBeenCalled();
+    expect(result.current.isError).toBe(false);
+    expect(result.current.data).toEqual(listPayload);
```
dtaylor113, line 95: Is this expect() just testing the the mock data 'listPayload' is being returned?

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
```diff
@@ -0,0 +1,349 @@
...
+    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } } as never);
+
+    const { result } = renderHook(() => useFetchClusterTransfer({ transferID: 't-sort' }));
+
+    await waitFor(() => expect(result.current.isLoading).toBe(false));
+
+    expect(searchClusterTransfersSpy).toHaveBeenCalledWith(
+      expect.objectContaining({ orderBy: 'updated_at desc' }),
+    );
+    expect(result.current.isError).toBe(false);
```
dtaylor113, line 139: Why is there no expect() test on the result?

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
```diff
@@ -0,0 +1,349 @@
...
+      }),
+    );
+    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } } as never);
+
+    const { result } = renderHook(() => useFetchClusterTransfer({ transferID: 't-asc' }));
+
+    await waitFor(() => expect(result.current.isLoading).toBe(false));
+
+    expect(searchClusterTransfersSpy).toHaveBeenCalledWith(
+      expect.objectContaining({ orderBy: 'name asc' }),
```
dtaylor113, line 164: If this is the only test, is it worth it?

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
```diff
@@ -0,0 +1,349 @@
...
+        type: viewPaginationConstants.SET_TOTAL_ITEMS,
+        payload: expect.objectContaining({
+          viewType: viewConstants.CLUSTER_TRANSFER_VIEW,
+          totalCount: 42,
+        }),
+      }),
+    );
+  });
+
+  it('returns pending transfer slice when showPendingTransfer is true', async () => {
```
dtaylor113: Pls elaborate on transfer 'slice'? - thanks

zherman0:
@dtaylor113 - Thanks for sending him the account info.  @LiorKGOW - I cannot put account credentials in a public PR.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
```diff
@@ -0,0 +1,349 @@
...
+    const { result } = renderHook(() =>
+      useFetchClusterTransfer({
+        clusterExternalID: 'ext-a',
+        showPendingTransfer: true,
+      }),
+    );
+
+    await waitFor(() => expect(result.current.isLoading).toBe(false));
+
+    expect(result.current.data).toEqual({ items: [acceptedItem] });
```
dtaylor113, line 282: Again, kinda confused.  We pass in acceptedItem to the mock/spy, then we just test if the function is returning the mock object?  I'm probably missing something.

Also, why "showPendingTransfer: true" in useFetchClusterTransfer(), but it returns 'accepted'?

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
```diff
@@ -0,0 +1,349 @@
...
+    );
+  });
+});
+
+describe('refetchFetchClusterTransfer', () => {
+  it('invalidates cluster transfer queries', () => {
+    const spy = jest.spyOn(queryClient, 'invalidateQueries');
+    refetchFetchClusterTransfer();
+    expect(spy).toHaveBeenCalledWith({
+      queryKey: [queryConstants.FETCH_CLUSTER_DETAILS_QUERY_KEY, 'fetchClusterTransfer'],
```
dtaylor113, line 345: How does this test 'invalidates cluster transfer queries'?

dtaylor113 (commented):
Hi @zherman0, I have some comments about the new unit test -thanks

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
zherman0: Yes, it is testing against the mockdata after it has gone though the entire data path for a single cluster.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
zherman0: In this case I don't check the data because the test is focused on the filter path; I did not think I needed to recheck the data path since we have done that elsewhere.  I did check the error to make sure the filter did not cause an issue.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
zherman0: So the other test is checking the default sort, this one is testing the explicit sort path.  These test are all about code coverage and I needed to test all paths to make my numbers, right?

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
zherman0: Slice is what golang calls a dynamic array.   I guess a better title would be 'returns pending transfer items'?

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
zherman0: Again, this was about code coverage.  For cluster transfers on the UI, we show pending and accepted items a certain way.  If you look at the code of the hook, you will see we check explicitly if the item is in the pending  or the accepted state, so this test is just making sure we cover that branch as well.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
zherman0: It's really testing that the query call is being made. I mocked out the actual react queries, so this is just checking that the call to invalidate on the queries object gets called when we call refetchFetchClusterTransfer.  This test seems silly and it is very simple, but it provides us code coverage and it verifies the refetch is successfully calling the invalidate.  But mainly this is just code test coverage - it is hard to test `invalidate` works in a unit test that is not wired into the API.  However, these lines of code count  towards the code coverage numbers - it is a tricky statistic to live by when a lot  of things are not meant to be tested in unit tests.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
dtaylor113: So this is testing mock object === mock object?  I that a useful test?

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
dtaylor113: Ok, thanks

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
dtaylor113: yes, pls update title to avoid confusion -thanks

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
dtaylor113: "...all about code coverage and I needed to test all paths to make my numbers, right?"
I'm not sure I agree.  Do we want light/pointless unit tests just for code coverage numbers?
Seems like an architect question, cc @kelvah  😄

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
zherman0: > So this is testing mock object === mock object?

Not at all. This is testing mockdata goes in representing data from the API then is goes through all the possible branches inside the hook, to then, for this particular test, to come out equaling the exact values that came in becasue that is what we expect for this use case.

>  [Is] that a useful test??

Yes.  It is testing that specific data path where we return all the results and no errors.  This covers a specific path through the code.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
zherman0: Again, this test is doing "something"; it is testing a specific path and yes it does help us reach our numbers.  I normally would not test the sorting in such detail, but there is only like 20 lines of code in this function so each line of unique logic represents like 5% on coverage, so I need to test everything.  Even if this seems pointless.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
dtaylor113: got it, thanks!

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
zherman0: Already updated

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
dtaylor113: Hmm...ok, would like others to weight in, my next two comments below are along these same lines/questions.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.tsx
LiorKGOW: > I see the confusion, this is not a feature request, this is a bug fix.
> 
> All the discussion is very valid, but can you please test/review this specific bug as the scope is very specific. Reach out to me for an ID to transfer to, as stated in the test instructions.

Precisely Zac! Specifically in this PR, it is a bug fix. I was referring to the entire feature which has been implemented several months ago. I wasn't reviewing only the fix. I investigated the fix with the context of the feature, to see if we are missing anything around it and see if there is a way to improve things as well in the process, as I haven't had a chance to touch this area of the code yet 😄 I believe in leaving a better code than the one we find  

As I noticed the discussion digressed from the scope of the PR, I suggested opening a ticket and addressing the findings on another PR 😇 I don't want to delay you from delivering the fix of course! 

I am sorry this dragged a little longer then expected. I got the credentials I need to test this PR locally, and plan to do it this morning. Thank you for your patience and cooperation regarding this matter Zac!

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
dtaylor113: Well, we didn't get any feedback and I don't feel it important enough to hold up this PR.
The unit-tests do test 'something', and if that's the only way to accurately reflect the code coverage numbers, I guess these types of unit tests are ok.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
dtaylor113: Ok with "hook checks explicitly if the item is in the pending or the accepted state, so this test is just making sure we cover that branch as well"

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
dtaylor113: Ok with " this is just checking that the call to invalidate on the queries object gets called when we call" and "it verifies the refetch is successfully calling the invalidate".

dtaylor113 (approved):
Tested thru 3 rounds of initiate -> cancel -> initiate, no errors.
code and new unit-test LGTM.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
```diff
@@ -0,0 +1,349 @@
...
+      selector({
+        viewOptions: {
+          [viewConstants.CLUSTER_TRANSFER_VIEW]: {
+            ...defaultViewOptions,
+            sorting: { ...defaultViewOptions.sorting, sortField: '' },
+          },
+        },
+      }),
+    );
+    searchClusterTransfersSpy.mockResolvedValueOnce({ data: { items: [], total: 0 } } as never);
```
LiorKGOW: I noticed the use of `as never` here, which is often used as an escape hatch to bypass strict type-checking. Could you share a bit more context on the reasoning behind using it in this case?

I’m also seeing several other occurrences of as never, which makes me wonder if there might be an underlying typing issue we could address more systematically.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
LiorKGOW, line 1: Thank you for adding the tests to this PR! I believe it would have been better to attach them with the PR adding this code originally, but better late then never 😊 Thank you for circling back to add them!

LiorKGOW: I believe @dtaylor113 's  suggestion was to add a unit test that covers a flow similar to the bug addressed in this PR. This would help ensure the issue is fully resolved and allow us to catch any similar regressions in the future.

Would it be possible to add such a test, @zherman0?

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
LiorKGOW: > "...all about code coverage and I needed to test all paths to make my numbers, right?"
> I'm not sure I agree. Do we want light/pointless unit tests just for code coverage numbers?
> Seems like an architect question, cc @kelvah 😄

@dtaylor113 if you will allow me to step in, 
First of all, aiming for at least 75% line coverage is something we previously found agreeable and achievable (throwback to [this Epic](https://redhat.atlassian.net/browse/OCMUI-3119)). This leads to a stable, (more) error prune project, and although not documented explicitly in our guidelines (specifically 75%, and maybe it should -> cc @kelvah ), it is something we should always aim for a 100% code coverage for new code entering the code base. The author of the code always knows how his code works, and is the best one to write them, which would also be helpful for others reading the code he wrote afterwards (seeing some examples of how the code is used,  in addition to understanding some of the edge cases), and even reuse it for other cases 😄 -> especially as we are talking about a hook

Additionally, if it is something worth examining and discussing, and the deadline is not tight, everything is worth bringing up for a discussion, especially if we are talking about concerns about the quality, readability, efficiency and stability of our application ! 😄 
The whole point of comments on PRs is having discussions, raising questions, for everybody to understand the code fully, and be sure of what enters the codebase, before it enters. 

Thank you for the thoroughness in this matter @zherman0 ! I just wanted to provide my opinion on this matter, and propose we anchor this in our guidelines (excuse me if it is already present in a similar way) @kelvah

LiorKGOW (commented):
> @dtaylor113 - Thanks for sending him the account info. @LiorKGOW - I cannot put account credentials in a public PR.

Of course I didn't expect you to post the credentials on a public PR @zherman0 😄

> Hi @LiorKGOW, I sent you Zac's Cluster Transfer info. that I used in a slack DM

Thank you @dtaylor113 ! I appreciate it 🤗

Although I couldn't reproduce the bug described on the how to test section of this PR (I couldn't get the error to pop out on the page, but saw a similar issue in the console when checked out on main), I did not see the error as you mentioned following the steps several times. Am I missing something? @zherman0 @dtaylor113 @shivanthi-amara Is there possibly something missing in the How to Test section? Couldn't find any other prerequisites to reproducing this bug 🤔

I don't want to delay you from merging this PR @zherman0, but I wanted to mention that I've encountered several errors on the console, when checked out out on the branch of this PR, which **were also present on the main branch** so **first**, I wanted to ask if you're encountering these errors as well @zherman0 @dtaylor113 @shivanthi-amara ?


Image

Image


**Second**, there was one error I noticed, which **was not present on main** but was present on your PR branch, which is somewhat relevant and worth looking into, as it is related to cluster transfers, and is a PATCH req which is the action we perform here. I suggest looking into it @zherman0 : 

Image


However, as I don't see anything wrong happening in the UI (from a user perspective), I think it would be ok to begin testing this PR, at least to try and reproduce the issue @shivanthi-amara mentioned on the ticket. 

With that, I prefer to submit my approval on a later time as I still have a few comments left to be addressed (additionally to ones I left on the test file). Thank you for your patience and thoroughness regarding this matter @zherman0 ! It is truly admirable ! 

I would also appreciate your opinion on some of the [comments](https://github.com/RedHatInsights/uhc-portal/pull/446/changes#r3118932437) I wrote previously, which you have yet to address 😊

zherman0:
@LiorKGOW 
Those errors you see are known about and generally they are extra calls being made on cluster list.  I cannot find the ticket, but we have discussed these in the past.

As for the cluster transfer error, it appears the rescind/cancel was pressed multiple times (which should not be possible) so I will look into that, although it is outside the scope of this PR.   I will create a new ticket if I can reproduce it - please let me know if you have repeatable steps to reproduce.

You mentioned I had not responded to your other comments - apologies - I looked through it and I could not find questions I did not address, would you mind re-asking it.  A lot of you writing is kind of stream-of-conscious and  I am not certain what expects a reply at times.

I will respond to the code changes in line.

I really appreciate your enthusiasm with this PR.  I will remind you the scope and fix was one line of code so I think we have gotten off track a little with this PR.   Please feel free to open any tickets you want (we can close them if they are dups).

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
zherman0: > Would it be possible to add such a test

Short answer: No.  Long Answer: Anything is possible..  So the issue here is that the updating of the react query data caused the key to change for the option views which then caused the dispatch to call redux which is what triggers the useEffect ... and then repeat.  With unit test, I mock all of the api calls and redux and dispatch, so I might be able to mock all of those items to return the exact conditions but it loses the timing aspect/real world interaction.  It would be complex and brittle.  If we were to try and make a test to imitate the original bug, I think an E2E test would make more sense.

zherman0: To clarify, an E2E test that duplicates the bug would be out of scope here.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
zherman0: Wow, just wow.  That is just AI slop that I did not ask for but was added somewhere along the line.  All removed and unneeded.

zherman0:
@LiorKGOW - With regards to reproducing the bug, I believe I copied the procedure from the jira ticket and I can always reproduce it in stage.  Let's touch base and go over it next week if you are having issues reproducing.

src/queries/ClusterDetailsQueries/ClusterTransferOwnership/useFetchClusterTransfer.test.tsx
LiorKGOW: I wouldn't say that this kind of test is not in scope here as, to the best of my knowledge, we usually add a test to test that the fix is ok test wise, so we won't need to worry about the bug resurfacing again 😄 

A unit test replicating the behavior mocking outside libraries would be required, but as this might be a special case of this behavior re-occurring occasionally, I'll drop this requirement this time. 
I'll appreciate if you could add one replicating the How to Test section just to be sure we get this right

LiorKGOW:
> @LiorKGOW Those errors you see are known about and generally they are extra calls being made on cluster list. I cannot find the ticket, but we have discussed these in the past.
> 
> As for the cluster transfer error, it appears the rescind/cancel was pressed multiple times (which should not be possible) so I will look into that, although it is outside the scope of this PR. I will create a new ticket if I can reproduce it - please let me know if you have repeatable steps to reproduce.
> 
> You mentioned I had not responded to your other comments - apologies - I looked through it and I could not find questions I did not address, would you mind re-asking it. A lot of you writing is kind of stream-of-conscious and I am not certain what expects a reply at times.
> 
> I will respond to the code changes in line.
> 
> I really appreciate your enthusiasm with this PR. I will remind you the scope and fix was one line of code so I think we have gotten off track a little with this PR. Please feel free to open any tickets you want (we can close them if they are dups).

I would love if you could point me out to those tickets, and if we can't find them, let's resurface them. 

Personally, I don't think we've fallen off track, as we have a good discussion and we both see the good of the project 😄

Regarding my questions that went un-answered, I'll appreciate your opinion:

1. Second, there was one error I noticed, which was not present on main but was present on your PR branch, which is somewhat relevant and worth looking into, as it is related to cluster transfers, and is a PATCH req which is the action we perform here. I suggest looking into it @zherman0 :

image

However, as I don't see anything wrong happening in the UI (from a user perspective), I think it would be ok to begin testing this PR, at least to try and reproduce the issue @shivanthi-amara mentioned on the ticket.

full comment context: https://github.com/RedHatInsights/uhc-portal/pull/446#pullrequestreview-4164974155 


2. It would be great if you could open the ticket and provide some context as you already worked in this area before and know what's going on there 😄
(Ex. Explain the flow, exactly what needs to be refactored out of redux - including all the relevant areas this is being used, if you considered alternatives)

I suggest to at least open a ticket and raising that to be worked on.
As we should also tackle tech debt equally as feature work, I would even consider prioritizing this, especially if we want to get rid of redux any time soon
cc @kelvah  

[full context of the comment](https://github.com/RedHatInsights/uhc-portal/pull/446/changes#r3118932437)

zherman0:
@LiorKGOW - I think we have gotten off track here.  This major bug that is in production today refers to a very specific error that is out in the wild.  That is the scope of this PR.  We are talking about re-writing pagination and removing redux on a bug.  That is why I think we are a little of topic.  I understand that everything is related and we can always improve things, but I am trying to prevent this issue currently.

zherman0:
@LiorKGOW regarding your  questions above:
> Second, there was one error I noticed, which was not present on main but was present on your PR branch, which is somewhat relevant and worth looking into, as it is related to cluster transfers, and is a PATCH req which is the action we perform here. 

I did respond to this.  I asked for steps to reproduce and I said I would open another ticket if I found a problem

> It would be great if you could open the ticket and provide some context as you already worked in this area before and know what's going on there

At this point, I don't think I will open a ticket to redo the pagination code.  That is a considerable change and I don't think it has the priority.  I will leave it to you or others if they want to open a ticket.

LiorKGOW (approved):
Focusing on the scope of this PR, I wasn’t able to reproduce the bug described in the ticket, and I don’t observe it occurring in your branch.

Regarding the PATCH request I mentioned earlier, I also wasn’t able to reproduce the issue, so I’m comfortable proceeding with this fix.

Regarding the surrounding code, I’ve discussed this with @kelvah , and we plan to bring it up with the team for further consideration, as the current approach raises some concerns.
Specifically, I’m not convinced there’s a clear benefit in using both Redux and React Query to manage the same piece of state, and it may be worth exploring a refactor.
I’d really appreciate your perspective on this as part of the broader discussion.

Thank you for your patience Zac

shivanthi-amara (approved):
Verified the changes and the issue cannot be reproduced anymore. LGTM

PR #447, opened 2026-04-12
OCMUI-902: HTPasswd IDP - support uploading an HTPasswd file

lizagilman:
# Summary

  Add support for uploading an HTPasswd file when creating an HTPasswd identity provider, as an alternative to manually entering users one by one.

  - New HTPasswdFileUpload component using PatternFly FileUpload
  - File parser (htpasswdFileParser.ts) with frontend validation (format, empty fields, duplicates)
  - Radio button toggle between "Add users manually" and "Upload an HTPasswd file"
  - Upload mode sends hashed_password key to the API; manual mode sends password as before
  - Yup validation skipped for user fields in upload mode (parser handles validation)
  - Unit tests for the parser

Day 2 (edit users list):
  - "Upload HTPasswd file" button added next to "Delete" and "Add user" in the HtpasswdDetails toolbar
  - Modal with file upload component, reuses the same parser/masking logic from Day 1
  - Bulk import via single POST to /htpasswd_users/import with { items: [{username, hashed_password}] }

  
  Assisted by: Claude Code / claude-opus-4-6

# Jira

  Fixes [OCMUI-902](https://redhat.atlassian.net/browse/OCMUI-902)

# Additional information

  - creationMode is stored in Formik values (not local state) so getCreateIDPRequestData() can access it to decide between password and hashed_password keys
  - Switching between manual and upload mode resets the users array and clears Formik touched state to prevent validation bleed
  - The parser and Yup validation are intentionally separate: parser validates on file upload, Yup validates manual entry fields
  - No changes needed to IDPForm.jsx or the API layer — upload populates the same users[] Formik field
  -  Day 2 adds createHtpasswdUserFromFile service method and useImportHtpasswdUsers mutation hook for the bulk import endpoint

Note on legacy Redux modal pattern:
`UploadHTPasswdFileModal` follows the existing Redux-based modal system used throughout `HtpasswdDetails`, which requires ts-ignore to satisfy TypeScript. This is against current project guidelines but was kept intentionally. All modals in this file must be migrated together to avoid a mixed approach.

# How to Test

  1. Navigate to a cluster's Identity Providers page
  2. Click "Add identity provider" and select HTPasswd
  3. Verify two radio buttons appear: "Add users manually" (default) and "Upload an HTPasswd file"
  4. Manual mode: works as before — add users with username/password fields
  5. Upload mode: upload a valid .htpasswd file (format: username:hashed_password per line)
   <strike> - Verify passwords are masked in the preview</strike> EDIT: This requirement was dropped 
    - Verify parser errors appear below the file upload for invalid files (missing colon, empty username/password, duplicates)
    - Verify the "Add" button enables after a valid file is uploaded
  6. Switch between modes and verify state resets cleanly
  7. In edit user flow, verify radio buttons are hidden and only manual mode shows
  
  Day 2 (edit the users list):
  1. Navigate to an existing HTPasswd IDP's details page
  2. Verify "Upload HTPasswd file" button appears next to "Delete" and "Add user"
  3. Click the button — a modal should open
  4. Upload a valid .htpasswd file — verify masked preview and Upload button enables
  5. Upload an invalid file — verify parser errors display and Upload button stays disabled
  6. Click Upload with a valid file — verify success notification and user list refreshes

Test file data:
```
user1:$2y$10$ABCDEFGHIJKLMNOPQRSTUUV0WXYZabcdefghijklmnopqrstuv
user2:$2y$10$1234567890ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijkl
user3:$apr1$xyz12345$ABCDEFGHIJKLMNOPQRSTUVw
```

To initiate an error:

```
user1:$2y$10$ABCDEFGHIJKLMNOPQRSTUUV0WXYZabcdefghijklmnopqrstuv
user2:$2y$10$1234567890ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijkl
:$apr1$xyz12345$ABCDEFGHIJKLMNOPQRSTUVw
```

# Screen Captures

 Day 1:

image

image

image

image

Day 2:

image

image

image

# Review process

## QE Reviewer

[OCMUI-902]: https://redhat.atlassian.net/browse/OCMUI-902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/HTPasswdForm.tsx
```diff
@@ -197,19 +214,56 @@ const HTPasswdForm = ({
...
-      onlySingleItem={onlySingleItem}
-    />
+    <>
+      {showModeSelection && (
+        <>
+          <Title headingLevel="h4" size="md">
+            Create users
+          </Title>
+          <Content component={ContentVariants.p} className="pf-v6-u-mb-0">
+            To creare the HTPasswd identity provider, you must create at least 1 user. You can add
```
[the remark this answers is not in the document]
lizagilman: fixed

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/HTPasswdFileUpload.tsx
```diff
@@ -0,0 +1,85 @@
...
+    setMaskedContent(maskHTPasswdFileContent(content));
+    setErrors(result.errors);
+
+    if (result.errors.length === 0 && result.users.length > 0) {
+      const formUsers = result.users.map((user) => ({
+        username: user.username,
+        password: user.password,
+      }));
+      setFieldValue(FieldId.USERS, formUsers);
+    }
```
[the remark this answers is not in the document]
lizagilman: Added else block to clear the stale users if a user uploads an invalid file after a valid one.

lizagilman:
For reviewers - note on legacy Redux modal pattern:
`UploadHTPasswdFileModal` follows the existing Redux-based modal system used throughout `HtpasswdDetails`, which requires  `ts-ignore` to satisfy TypeScript. This is against current project guidelines but was kept intentionally. All modals in this file must be migrated together to avoid a mixed approach.

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  React.useEffect(() => {
+    if (isSuccess) {
+      onSuccess();
+      closeUploadModal();
+      addNotification({
+        variant: 'success',
+        title: `Successfully imported ${parsedUsers.length} user${parsedUsers.length > 1 ? 's' : ''}`,
+        dismissable: true,
+      });
```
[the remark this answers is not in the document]
lizagilman: 0 users is not a real scenario, test updated.

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+    withState(initialState, true).render(<UploadHTPasswdFileModal onSuccess={onSuccess} />);
+
+    expect(reset).toHaveBeenCalled();
+    expect(mockedDispatch.mock.calls[0][0].type).toEqual('CLOSE_MODAL');
+    expect(onSuccess).toHaveBeenCalled();
+    expect(mockedAddNotification).toHaveBeenCalledWith({
+      dismissable: true,
+      title: 'Successfully imported 0 user',
+      variant: 'success',
+    });
```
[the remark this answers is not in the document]
lizagilman: 0 users is not a real scenario, test updated.

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/IDPForm.jsx
```diff
@@ -154,9 +155,15 @@ const IDPForm = ({
         text = (
           <>
             Define an <code>htpasswd</code> identity provider for your managed cluster to create one
-            or multiple static users that can log in to your cluster and troubleshoot it. If these
-            users need elevated permissions, add it to an{' '}
-            <ExternalLink href={docLinks.OSD_DEDICATED_ADMIN_ROLE}>
+            or multiple static users that can log in to your cluster. If these users need elevated
```
eliranmal, line 165: is omitting _'and troubleshoot it'_ here intentional?

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/IDPForm.jsx
```diff
@@ -154,9 +155,15 @@ const IDPForm = ({
...
+            permissions, add them to{' '}
```
eliranmal: we should retain the _'an'_ after _'add them to'_.

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/IDPForm.jsx
```diff
@@ -154,9 +155,15 @@ const IDPForm = ({
...
+            <ExternalLink
+              href={
+                isROSACluster
+                  ? docLinks.ROSA_AUTHENTICATION_AND_AUTHORIZATION
+                  : docLinks.OSD_DEDICATED_ADMIN_ROLE
+              }
+            >
               administrative group
```
eliranmal: i couldn't find the term _"administrative group"_ in the docs - they talk about _"administration group"_ instead.

maybe we should update this while we're at it.

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/IDPForm.jsx
```diff
@@ -154,9 +155,15 @@ const IDPForm = ({
...
             Define an <code>htpasswd</code> identity provider for your managed cluster to create one
-            or multiple static users that can log in to your cluster and troubleshoot it. If these
-            users need elevated permissions, add it to an{' '}
-            <ExternalLink href={docLinks.OSD_DEDICATED_ADMIN_ROLE}>
+            or multiple static users that can log in to your cluster. If these users need elevated
+            permissions, add them to{' '}
+            <ExternalLink
+              href={
+                isROSACluster
+                  ? docLinks.ROSA_AUTHENTICATION_AND_AUTHORIZATION
```
eliranmal: should the ROSA link point to the relevant page section?

i.e. section _"4.7. Configuring an htpasswd identity provider"_, via the anchor link `#config-htpasswd-idp_sd-configuring-identity-providers`.

src/common/docLinks.mjs
```diff
@@ -108,6 +109,7 @@ const docLinks = {
   // OCP
   AWS_CONTROL_PLANE_URL: `${OCP_DOCS_BASE}/architecture/control-plane#control-plane`,
   IDP_HTPASSWD: `${OCP_DOCS_BASE}/authentication_and_authorization/configuring-identity-providers#configuring-htpasswd-identity-provider`,
+  IDP_HTPASSWD_UTILITY: `https://docs.redhat.com/en/documentation/openshift_container_platform/4.8/html-single/authentication_and_authorization/index?extIdCarryOver=true&sc_cid=RHCTG0180000371695#creating-htpasswd-file`,
```
eliranmal: the query string is redundant; please remove it.

(i.e. the `?extIdCarryOver=true&sc_cid=RHCTG0180000371695` part)

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+import * as useImportHtpasswdUsersModule from '~/queries/ClusterDetailsQueries/AccessControlTab/UserQueries/useImportHtpasswdUsers';
+import { screen, userEvent, waitFor, withState } from '~/testUtils';
+
+import UploadHTPasswdFileModal from './UploadHTPasswdFileModal';
+
+jest.mock('react-redux', () => ({
+  __esModule: true,
+  ...jest.requireActual('react-redux'),
+}));
```
eliranmal, line 15: hmm, a mock with only `requireActual` in it, but no extension of any method makes no sense to me..

isn't this just a no-op?

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+const mockedAddNotification = jest.fn();
+
+jest.mock('@redhat-cloud-services/frontend-components-notifications', () => ({
+  __esModule: true,
+  ...jest.requireActual('@redhat-cloud-services/frontend-components-notifications'),
```
eliranmal: why do we need `requireActual` here?

the test-cases don't reference any of the other methods in `frontend-components-notifications` (aside from the mocked `useAddNotification`) 🤔

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+      idpName: 'myIDPName',
+      clusterId: 'myClusterId',
+      idpId: 'myIDPID',
+      region: 'us-east-1',
+    },
+  },
+};
+
+const validFileContent = 'user1:$2y$05$hash1\nuser2:$2y$05$hash2';
+const invalidFileContent = 'invalidline\nuser1:pass1';
```
eliranmal: - nit: for readability, please use template-literals and separate lines with actual newlines, instead of using delimiters (`\n`)

- can we add more scenarios for invalid files?  i.e.:
  - a single line with just colons and password (`:<hashed-pass>`)
  - same but with the name (`<name>:`)
  - just colons
  - an empty file
  - a file with the wrong extension  

  (the top three are probably nit-picks, but the last two are important IMO)

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+const uploadFile = async (content: string, filename = 'users.htpasswd') => {
+  const file = new File([content], filename, { type: 'text/plain' });
+  const fileInput = document.querySelector('input[type="file"]') as HTMLInputElement;
+  await userEvent.upload(fileInput, file);
+};
+
```
eliranmal: nit:

this fragment is both duplicated in this PR, and duplicates an existing pattern in the code (see [_ReduxFileUpload.test.jsx_ lines 122-125][1]).

can we extract it to a common utility somewhere (maybe _testUtils.tsx_)?


[1]: https://github.com/RedHatInsights/uhc-portal/blob/1bac91f60d5cdbe4a054bb2857c9dfaf9e0eb99c/src/components/common/ReduxFormComponents_deprecated/__tests__/ReduxFileUpload.test.jsx#L122-L125

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+  it('calls mutate with parsed users when Upload is clicked', async () => {
+    mockedImportUsers.mockReturnValue(defaultReturn);
+
+    const { user } = withState(initialState, true).render(
+      <UploadHTPasswdFileModal onSuccess={onSuccess} />,
+    );
+
+    await uploadFile(validFileContent);
+
+    await waitFor(() => {
```
eliranmal, line 181: nit:

we should be able to use the more succinct `queryBy`/`not` pattern here (which is also sync).

i.e. instead of:
```
await waitFor(expect(getBy()).not.toBe())
```
use:
```
expect(queryBy()).not.toBe()
```

(same goes for other occurrences in the PR)

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+    clusterId,
+    idpId,
+    region,
+  );
+
+  const [filename, setFilename] = React.useState('');
+  const [parseErrors, setParseErrors] = React.useState<string[]>([]);
+  const [parsedUsers, setParsedUsers] = React.useState<ParsedHTPasswdUser[]>([]);
+
+  const closeUploadModal = React.useCallback(() => {
```
eliranmal: nit:  why not import `useState`, `useCallback` etc. directly?  no need for `React.` here.

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+    dispatch(closeModal());
+  }, [dispatch, reset]);
+
+  React.useEffect(() => {
+    if (isSuccess) {
+      onSuccess();
+      closeUploadModal();
+      addNotification({
+        variant: 'success',
+        title: `Successfully imported ${parsedUsers.length} user${parsedUsers.length > 1 ? 's' : ''}`,
```
eliranmal: possible bug: `0` is also plural.

i realize `0` isn't expected here so this is maybe a nit, but for correctness, we might wanna update to `length === 1 ? '' : 's'`.

(btw, there's no current pattern around this – we don't seem to have a utility function for that, and implementations differ across the codebase).

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+        dismissable: true,
+      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [closeUploadModal, isSuccess, onSuccess]);
```
eliranmal: should `parsedUsers` be included in the deps array?

(or  `parsedUsers.length`)

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  const processContent = (content: string) => {
+    const result = parseHTPasswdFile(content);
+    setParseErrors(result.errors);
+    setParsedUsers(result.errors.length === 0 ? result.users : []);
```
eliranmal, line 66: i'd expect this should be handled by `parseHTPasswdFile()`; i.e. if there are errors, `users` should be empty.

(this will also simplify other usages, like `parsedUsers.length === 0 || parseErrors.length > 0` in [line 96][1])


[1]: https://github.com/RedHatInsights/uhc-portal/pull/447/changes#diff-cabf4d589472ab3cb0a9b9c1f15f36d90da4ab45483f19aa6b3b4b71060c5100R96

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+    >
+      <Form>
+        <p>
+          Upload a valid htpasswd file to add users to identity provider <strong>{idpName}</strong>.
+          Generally, this file is prepared using the{' '}
+          <a href={docLinks.IDP_HTPASSWD_UTILITY} target="_blank" rel="noreferrer">
+            htpasswd
+          </a>{' '}
+          tool. Each line must contain a username and a hashed password. If any user fails to be
+          created for any reason, the entire import is cancelled — no users will be added.
```
eliranmal, line 116: nit:  this excerpt is mainly duplicated (aside from a word or two) – should we extract it to a utility function?

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/htpasswdUtilities.ts
```diff
@@ -3,4 +3,4 @@ import { HtPasswdIdentityProvider } from '~/types/clusters_mgmt.v1';
 export const isSingleUserHtpasswd = (htpasswd: HtPasswdIdentityProvider) => !!htpasswd?.username;
 
 export const singleUserHtpasswdMessage =
-  'Single user HTPasswd IDPs cannot be modified. Delete the IDP and recreate it as a multi user HTPasswd IDP.';
+  'Single user htpasswd IDPs cannot be modified. Delete the IDP and recreate it as a multi user HTPasswd IDP.';
```
meirinetta: Should the second 'HTPasswd' also be with no capitalization?

src/common/docLinks.mjs
```diff
@@ -80,6 +80,7 @@ const docLinks = {
 
   // OSD
   OSD_DEDICATED_ADMIN_ROLE: `${OSD_DOCS_BASE}/authentication_and_authorization/osd-admin-roles`,
+  ROSA_AUTHENTICATION_AND_AUTHORIZATION: `${ROSA_DOCS_BASE}-single/authentication_and_authorization/index`,
```
meirinetta: Since this doc link exists in both classic and hcp clusters, I think this should be split to separate links.
We had a few months ago a [PR](https://github.com/RedHatInsights/uhc-portal/pull/280) to separate doc links for classic and hcp.
It looks like the link for classic is - https://docs.redhat.com/en/documentation/red_hat_openshift_service_on_aws_classic_architecture/4/html/authentication_and_authorization/index
but it's better to check with the docs team about this link.

meirinetta (commented):
Hey @lizagilman, tested locally with valid and invalid scenarios, all behaved correctly. Added 2 small comments.

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/htpasswdUtilities.ts
lizagilman: Fixed, good catch, thanks!

src/common/docLinks.mjs
lizagilman: Checking with PM

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/HtpasswdDetails.tsx
```diff
@@ -49,6 +52,7 @@ const HtpasswdDetails = ({
   isSingleUserHtpasswd,
 }: Props) => {
   const dispatch = useDispatch();
+  const isImportEnabled = useFeatureGate(HTPASSWD_IMPORT);
```
eliranmal, line 55: nit: please rename to `isHtpasswdImportEnabled` to avoid ambiguity.

(same goes for other occurrences)

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/HtpasswdDetails.test.tsx
```diff
@@ -644,6 +652,40 @@ describe('<HtpasswdDetails />', () => {
     }, 20000);
   });
 
+  describe('Upload htpasswd file button', () => {
+    it('is not shown when feature flag is off', () => {
```
eliranmal, line 656: please explicitly set the feature-gate to 'off', to ensure we test the intended behavior (and not rely on any fallback value).

this is a pattern already in use in the codebase.

(same goes for other similar occurrences)

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.test.ts
eliranmal, line 1: same comment as earlier: in fixtures, please use template-literals with actual line-breaks instead of `\n`, for readability.

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.ts
```diff
@@ -0,0 +1,56 @@
+export type ParsedHTPasswdUser = {
+  username: string;
+  password: string;
+};
+
+export type HTPasswdParseResult = {
+  users: ParsedHTPasswdUser[];
```
eliranmal, line 7: nit:  this is not just a user - maybe rename 'user' to 'entry' in the field name and type for better semantics.

i.e.:  `entries: ParsedHTPasswdEntry[]`

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.ts
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+
+    if (seenUsernames.has(username)) {
+      errors.push(`Line ${index + 1}: Duplicate username "${username}".`);
+      return;
+    }
+
+    seenUsernames.add(username);
+    users.push({ username, password });
```
eliranmal, line 57: these 2 updates are a redundancy; either:

- remove `seenUsernames` – the same item can be retrieved from `users[last].username`, and there's no use for a set here because the control-flow never actually reaches a duplicate item (the loop quits before that)
- refactor to use a `map()` instead of `forEach()`, return the `{username, password}` entry instead of pushing to `users`, and move `users` outside the loop to be initialized with the return value of `map()` (if we still want to use 2 separate vars just for readability)

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.ts
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  if (users.length === 0 && errors.length === 0) {
```
eliranmal, line 60: strictly speaking, `errors` state isn't related to this condition - i think it can be omitted.

(cuz more code == more bugs :/ )

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/HTPasswdFileUpload.tsx
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  const onDataChange = (_event: DropEvent, content: string) => {
+    processContent(content);
+  };
+
+  const onClearClick = () => {
+    setFilename('');
+    setErrors([]);
+    setFieldValue(FieldId.USERS, [{ username: '', password: '' }]);
```
eliranmal: nit: we can use an empty array to reset the field  (if it doesn't require too much adapting to make related code more defensive)

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/HTPasswdForm.tsx
```diff
@@ -141,7 +154,21 @@ const HTPasswdForm = ({
   isEdit?: boolean;
   user?: any;
 }) => {
-  const { getFieldMeta, setFieldValue } = useFormState();
+  const { getFieldMeta, setFieldValue, setTouched, touched, values } = useFormState();
+  const isImportEnabled = useFeatureGate(HTPASSWD_IMPORT);
+  const creationMode: CreationMode = isImportEnabled
+    ? values[FieldId.CREATION_MODE] || CREATION_MODE_MANUAL
+    : CREATION_MODE_MANUAL;
```
eliranmal, line 161: this might hide some bugs - the fallback value only appears in this assignment, but not when setting the field value elsewhere.  
it also makes the condition cumbersome because of the repetition of `CREATION_MODE_MANUAL`.

we should probably move the fallback value of `FieldId.CREATION_MODE` to the form's initial values, and to any field initialization (e.g. in `handleModeChange`), and then simplify the condition to:

```
const creationMode: CreationMode = isImportEnabled
    ? values[FieldId.CREATION_MODE]
    : CREATION_MODE_MANUAL;
```

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/HTPasswdForm.tsx
```diff
@@ -197,19 +223,69 @@ const HTPasswdForm = ({
...
+              label="Add users manually"
+              id="htpasswd-mode-manual"
+            />
+            <Radio
+              isChecked={creationMode === CREATION_MODE_UPLOAD}
+              name="htpasswd-creation-mode"
+              onChange={() => handleModeChange(CREATION_MODE_UPLOAD)}
+              label="Upload an htpasswd file"
+              id="htpasswd-mode-upload"
+            />
```
eliranmal, line 248: `id`s might be redundant when `name` is stated as field identifier for the form

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/constants.ts
```diff
@@ -32,4 +32,9 @@ export enum FieldId {
   GITHUB_AUTH_MODE = 'github_auth_mode',
   ORGANIZATIONS = 'organizations',
   TEAMS = 'teams',
+  CREATION_MODE = 'creationMode',
 }
+
+export const CREATION_MODE_MANUAL = 'manual';
+export const CREATION_MODE_UPLOAD = 'upload';
+export type CreationMode = 'manual' | 'upload';
```
eliranmal: please reuse constants, i.e. update to:

```
export type CreationMode = CREATION_MODE_MANUAL | CREATION_MODE_UPLOAD;
```

eliranmal (changes requested):
behavior seems ok, i had some comments on the code - see inline ✏️ 

additionally:

- "how to test" section says: _"Verify passwords are masked in the preview"_.  what preview?  when uploading a file there's no preview where passwords can be seen 🤔 
- the upload field allows attaching a file of the wrong type, which result in a bunch of meaningless/unuseful errors (see screenshot).  we should filter the upload field by file type, to only allow the `.htpasswd` extension

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/constants.ts
lizagilman: Type expects type(/types), not values. `CREATION_MODE_MANUAL` and `CREATION_MODE_MANUAL` are values. But -
```
export type CreationMode =                                                                                                                                                                                                                  
    | typeof CREATION_MODE_MANUAL                                                                                                                                                                                                             
    | typeof CREATION_MODE_UPLOAD; 
````
should do the trick.

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/IDPForm.jsx (hunk 5 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/IDPForm.jsx:165]: Yes, PM asked to removed it.

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/constants.ts (hunk 30 above)
Comment (eliranmal) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/constants.ts]: oh got ya, whatever works really 👌    
it just looked like a DRY opportunity

File: src/common/docLinks.mjs (hunk 21 above)
Comment (lizagilman) [reply src/common/docLinks.mjs]: Updated, thanks

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/IDPForm.jsx (hunk 6 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/IDPForm.jsx]: Good catch fixed thanks

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/IDPForm.jsx (hunk 7 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/IDPForm.jsx]: This copy suggested and approved by PM, I'd leave it as is. The meaning is clear.

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.test.tsx (hunk 10 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.test.tsx:15]: This is need to create the mock for `useDispacth`

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.tsx (hunk 16 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.tsx]: updated but yeah this is never expected to happen, empty file will not pass frontend and backend valiation.

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/HtpasswdDetails.tsx (hunk 22 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/HtpasswdDetails.tsx:55]: Since this is within the scope of the Htpasswd IDP (both in terms of file structure and conceptually), I think it’s reasonably clear that it refers to enabling importing htpasswd. Given that this is a minor nit, I"ll leave it as is.

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.ts (hunk 24 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.ts:7]: Conceptually it refers to users (with passwords, right, but still it describes users). Since it's a nit, I"ll leave it as is.

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.test.tsx (hunk 10 above)
Comment (eliranmal) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.test.tsx:15]: ok 🤷

**Comment by lizagilman** [comment]
> behavior seems ok, i had some comments on the code - see inline ✏️
> 
> additionally:
> 
> * "how to test" section says: _"Verify passwords are masked in the preview"_.  what preview?  when uploading a file there's no preview where passwords can be seen 🤔
> * the upload field allows attaching a file of the wrong type, which result in a bunch of meaningless/unuseful errors (see screenshot).  we should filter the upload field by file type, to only allow the `.htpasswd` extension
> 


The masked passwords is an outdated requirement, I updated the description. Sorry for the confusion.
Adding restrictions on file types (.htpasswd and .txt ).

**Comment by lizagilman** [comment]
Added file type validation:

image

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.ts (hunk 26 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.ts:60]: Without this the empty file validation error will be shown, which is not correct. It should be shown only when the file is actually empty.

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/HTPasswdFileUpload.tsx (hunk 27 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/HTPasswdFileUpload.tsx]: It doesn't make much difference but updated for consistency.

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/HTPasswdForm.tsx (hunk 28 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/HTPasswdForm.tsx:161]: updated

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.ts (hunk 25 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.ts:57]: The loop doesn't quit, it continues. `seenUsernames` is to detect duplicate users.

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.test.ts
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.test.ts:1]: Too many nitpicks, if this would be the only one I'd address it, but I'd rather focus on the major stuff.

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.tsx (hunk 19 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.tsx:116]: Given:
- Conceptually these are 2 separate things - Day 1 and Day 2.
- They are not 100% the same
- This is one of many nitpicks and I'd rather focus on other more important stuff

I'm leaving it as is. But thanks for the suggestion.

File: src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.test.tsx (hunk 14 above)
Comment (lizagilman) [reply src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/HtpasswdDetails/UploadHTPasswdFileModal.test.tsx:181]: Not sure it will work, the button state changes asynchroniusly, we need to wait on it. Skipping since this is a nit. Thanks

File: src/common/docLinks.mjs [hunk 31]
```diff
@@ -80,6 +80,8 @@ const docLinks = {
 
   // OSD
   OSD_DEDICATED_ADMIN_ROLE: `${OSD_DOCS_BASE}/authentication_and_authorization/osd-admin-roles`,
+  ROSA_HCP_AUTHENTICATION_AND_AUTHORIZATION: `${ROSA_DOCS_BASE}-single/authentication_and_authorization/index#config-htpasswd-idp_sd-configuring-identity-providers`,
+  ROSA_CLASSIC_AUTHENTICATION_AND_AUTHORIZATION: `${ROSA_CLASSIC_DOCS_BASE}-single/authentication_and_authorization/index#config-htpasswd-idp_sd-configuring-identity-providers`,
```
eliranmal: nit:  these links no longer point to the auth' page - they lead to an inner section.

please rename them to reflect their semantics:

```
ROSA_HCP_AUTH_HTPASSWD_CONFIG
ROSA_CLASSIC_AUTH_HTPASSWD_CONFIG
```

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/HTPasswdForm.test.jsx
```diff
@@ -72,7 +74,43 @@ describe('HTPasswdForm', () => {
     expect(await screen.findByRole('button', { name: 'Add user' })).toBeDisabled();
   });
 
+  it('does not show mode selection radio buttons when feature flag is off', () => {
```
eliranmal, line 77: [this comment][1] called for this change throughout the PR:

> please explicitly set the feature-gate to 'off', to ensure we test the intended behavior (and not rely on any fallback value).
> 
> this is a pattern already in use in the codebase.
> 
> (same goes for other similar occurrences)

... but this occurrence was left out.  please update it too.


[1]: https://github.com/RedHatInsights/uhc-portal/pull/447#discussion_r3148186854

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.ts
eliranmal: yes, i've seen that, my points still hold.

`users` can also be used to detect duplicates, for example with:

```
users.find(u => u.username === username)
```

---

this is not a blocker tho, so feel free to ignore this.

eliranmal (commented):
tested the new behavior of the file-type, seems to be working fine 👌 

had a few minor comments – see inline ✏️ 

CI is failing for some reason, please have a look 🔎

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/HTPasswdForm.test.jsx
lizagilman: I fixed it - https://github.com/RedHatInsights/uhc-portal/pull/447/changes#diff-1ce620cf0c9c6f3098c56e91c071b32f6cf3ea4d0a0fe5bd9996e36c49f3218cR657

src/common/docLinks.mjs
lizagilman: done

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/htpasswdFileParser.ts
lizagilman: But I still have to store them, no? 
Let's leave as is :)

src/components/clusters/ClusterDetailsMultiRegion/components/IdentityProvidersPage/components/ProvidersForms/HTPasswdForm.test.jsx
eliranmal: yes, you fixed that other occurrence, but i'm saying that comment referred to all occurrences, and this one here is still lacking.

to clarify - we also need to add this here, too:

```
mockUseFeatureGate([[HTPASSWD_IMPORT, false]]);
```

lizagilman: whoops, missed that! Thanks!

eliranmal (approved):

meirinetta (approved):
Thanks for the changes, LGTM

jmekkatt:
Hi @lizagilman , "Add" button wrongly enabled after changing the name of the htpasswd idp and without uploading the required password file.

https://github.com/user-attachments/assets/b89ec244-5d32-4ca2-96a9-a57a31c9ca07

jmekkatt:
> Hi @lizagilman , "Add" button wrongly enabled after changing the name of the htpasswd idp and without uploading the required password file.
> 
>  Screen.Recording.2026-05-04.at.5.29.46.PM.mov

Add button also enabled wrongly  in below circumstances.
1. A user kept the username , both password fields under "Add users manually" section and switch to "Upload an htpasswd file" - Add button wrongly enabled without htpasswd file.
2. User uploaded wrongly formatted htpasswd  file from "Upload an htpasswd file" section

jmekkatt:
@lizagilman , There are character restrictions in place for the  user names and password formats while adding users via "Add users manually" but the same restrictions are ignored during the "Upload an htpasswd file" option.

The password file contains following data
1. The username contains white space and unsupported special characters
2. The password doesn't satisfy the minimum UI requirement shown in "Add users manually" 
Example,
```
test:%testing:sam
Test ing:hi
```
The password file upload seems successful without any error and some of the special characters were omitted after user creation.
See the recording

https://github.com/user-attachments/assets/ac2468ff-e60e-4d33-817e-d6f3faff311c

jmekkatt:
Hi @lizagilman , The API error on user import failure stays on the upload htpasswd file model even after user clears the selected password file. Expectation is that , once user clears the selected password file then the API error associated to the user import failure should be cleared from the model.
See the recording attached.
https://github.com/user-attachments/assets/bfb7c023-4bba-422e-8d74-07d38a68651a

lizagilman:
> Hi @lizagilman , The API error on user import failure stays on the upload htpasswd file model even after user clears the selected password file. Expectation is that , once user clears the selected password file then the API error associated to the user import failure should be cleared from the model. See the recording attached. https://github.com/user-attachments/assets/bfb7c023-4bba-422e-8d74-07d38a68651a


Hi Jaya this is intentional, I was thinking if user only cleared but doesn't yet select a new file then maybe it's a good idea to let them still be aware of what went wrong and only clear the error after the actually pick a new file (clicking the browse button). I consulted with UX and the feedback I got is that it's better to get rid of the clear button since it doesn't matter much and keep the error till they pick a new file. So I will try to remove it, otherwise I"ll keep it as is. Unless, we have a clear pattern for that, that we should hide errors on clear button click? Not sure about it, not that I'm aware of.

**Edit**: I think I"ll keep the clear button, to keep consistency across the app (other file upload components has it). I would leave this error as is, cleared on file change, not on "clear" click. OK?

jmekkatt:
Hi @lizagilman , 
The selected htpasswd file is not preserved when the user navigates away from the upload dialog (e.g., by clicking the documentation link) and returns. This leads to a poor user experience, as the user must reselect the file. The expectation is to retain the dialog state during such navigation, and triggering a full page refresh in this flow seems unnecessary and disruptive.

Please refer the attached recordings

https://github.com/user-attachments/assets/15e2d213-08a1-4f4c-a118-8dd3a9efebbb

jmekkatt:
Hi @lizagilman , The system currently allows creation of an HTPasswd IDP using a file with plain text passwords. Although the IDP is created successfully, authentication fails and no users or identities appear in the cluster, which leads to confusion. It would be better to validate the password format during upload (or at least warn the user) to prevent creating a non-functional IDP. Could you please check the same?
Example password file
```
test:JKN@#$123
```

jmekkatt (changes requested):
Hi @lizagilman , Kindly review the findings when you get a chance. Thank you.

lizagilman:
@jaya Hi, about - 

> 1. @lizagilman , There are character restrictions in place for the user names and password formats while adding users via "Add users manually" but the same restrictions are ignored during the "Upload an htpasswd file" option.
> 
> The password file contains following data
> 
> The username contains white space and unsupported special characters
> The password doesn't satisfy the minimum UI requirement shown in "Add users manually"
> Example,
> test:%testing:sam
> Test ing:hi
> The password file upload seems successful without any error and some of the special characters were omitted after user creation.

And 
2. 

> Hi @lizagilman , The system currently allows creation of an HTPasswd IDP using a file with plain text passwords. Although the IDP is created successfully, authentication fails and no users or identities appear in the cluster, which leads to confusion. It would be better to validate the password format during upload (or at least warn the user) to prevent creating a non-functional IDP. Could you please check the same?
Example password file

test:JKN@#$123

If no validation error from backend, these are backend bugs. 

This is (wrongly) valid request -

```

{
	"items": [
		{
			"hashed_password": "%testing:sam",
			"username": "test"
		},
		{
			"hashed_password": "hi",
			"username": "Test ing"
		}
	]
}

```


About - `The password doesn't satisfy the minimum UI requirement shown in "Add users manually"` - it shouldn't rely on the same validation.As you have also probably noticed, importing from files expects hashed passwords, and manual creation passwords should be plain text passwords. I"ll open a ticket to backend team. 

Anyhow, adding frontend validation.

lizagilman:
@jmekkatt I believe I addressed all issues and comment, PTAL. Thanks.

jmekkatt:
Hi @lizagilman , I have completed pre-merge testing for this change and changes looks good to me.
One last thing, Do you know why do we point the htpasswd tool doc link to OCP 4.8 or older version ex: https://docs.redhat.com/en/documentation/openshift_container_platform/4.8/html-single/authentication_and_authorization/index#creating-htpasswd-file ? Could we point to the latest OCP version doc? 
Screenshot 2026-05-11 at 3 28 34 PM

jmekkatt (approved):
Pre-merge testing has done and LGTM.

PR #455, opened 2026-04-15
OCMUI-4200 - [GCP OSD wizard]Service Account JSON validation always shows incorrect "requires osd-ccs-admin" error for client_email format issues

DcooperRH:
# Summary
This PR adds a fix to the validation function for the GCP service account json/file upload component. Previously if a user entered an incorrect @ domain in the client_email field, the validation would read `"The provided JSON does not meet the requirements: The field 'client_email' requires a service account name of 'osd-ccs-admin'."` This error message is unclear to the user of what is actually wrong in the format of client_email. Now the validation reads `"The provided JSON does not meet the requirements: The field 'client_email' must be a Google Cloud service account email ending in '.iam.gserviceaccount.com'."` 

# Jira

Fixes [OCMUI-4200](https://issues.redhat.com/browse/OCMUI-4200) 

# How to Test

1. Begin creating an OSD GCP CCS cluster
2. Select authentication as service account.
3. Locate the "Service account JSON" file upload field
4. Upload (or paste) a valid JSON structure where client_email has the correct service account name (osd-ccs-admin) but an incorrect domain format, for example:"client_email": "osd-ccs-admin@wrong-domain.com",
5. Check that the proper validation message displays 

 

# Screen Captures

| ocmui4200_before | ocmui4200_after |

# Review process

## QE Reviewer

[OCMUI-4200]: https://redhat.atlassian.net/browse/OCMUI-4200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/common/validators.ts
+++ b/src/common/validators.ts
@@ -1414,13 +1414,24 @@ const validateGCPServiceAccount = (content: string): string | undefined => {
       return 'Invalid JSON format.';
     }
     if (e instanceof ValidationError) {
-      let errorMessage;
+      let errorMessage: string;
       if (e.property.startsWith('instance.')) {
         const errorFieldName = e.property.replace('instance.', '');
-        if (errorFieldName === 'client_email' && e.instance.split[0] !== 'osd-ccs-admin') {
+        const clientEmailLocalPart =
+          errorFieldName === 'client_email' && typeof e.instance === 'string'
+            ? e.instance.split('@')[0]
+            : undefined;
+        if (
+          errorFieldName === 'client_email' &&
+          clientEmailLocalPart !== undefined &&
+          clientEmailLocalPart !== 'osd-ccs-admin'
+        ) {
           errorMessage = `The field '${errorFieldName}' requires a service account name of 'osd-ccs-admin'.`;
         } else if (e.message.indexOf('does not match pattern') !== -1) {
-          errorMessage = `The field '${errorFieldName}' is not in the required format.`;
+          errorMessage =
+            errorFieldName === 'client_email' && clientEmailLocalPart === 'osd-ccs-admin'
+              ? `The field '${errorFieldName}' must be a Google Cloud service account email ending in '.iam.gserviceaccount.com'.`
+              : `The field '${errorFieldName}' is not in the required format.`;
         } else {
           errorMessage = `The field '${errorFieldName}' ${e.message}`;
         }
--- a/src/common/__tests__/validators.test.ts
+++ b/src/common/__tests__/validators.test.ts
@@ -25,6 +25,7 @@ import validators, {
   required,
   validateAWSKMSKeyARN,
   validateGCPKMSServiceAccount,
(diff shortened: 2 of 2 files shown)


jmekkatt (approved):
Changes lgtm

Image

meirinetta (approved):
Changes LGTM

src/common/validators.ts
```diff
@@ -1414,13 +1414,24 @@ const validateGCPServiceAccount = (content: string): string | undefined => {
...
       if (e.property.startsWith('instance.')) {
         const errorFieldName = e.property.replace('instance.', '');
-        if (errorFieldName === 'client_email' && e.instance.split[0] !== 'osd-ccs-admin') {
+        const clientEmailLocalPart =
+          errorFieldName === 'client_email' && typeof e.instance === 'string'
+            ? e.instance.split('@')[0]
+            : undefined;
+        if (
+          errorFieldName === 'client_email' &&
+          clientEmailLocalPart !== undefined &&
```
LiorKGOW, line 1426: Non-blocking 😄
Generally, I would go for a null value, as undefined might be unexpectedly undefined for some other reason, but for a null value, we know it it should be null and are expecting it (talking about clientEmailLocalPart)

If we are talking about an error message, and we don;t have an error, there is no real reason to expect a defined error message, so I guess my claim does not stand in this specific case, but as we are talking about a value inside the email, I would go for null instead of undefined

src/common/validators.ts
```diff
@@ -1414,13 +1414,24 @@ const validateGCPServiceAccount = (content: string): string | undefined => {
...
+          clientEmailLocalPart !== 'osd-ccs-admin'
+        ) {
           errorMessage = `The field '${errorFieldName}' requires a service account name of 'osd-ccs-admin'.`;
         } else if (e.message.indexOf('does not match pattern') !== -1) {
-          errorMessage = `The field '${errorFieldName}' is not in the required format.`;
+          errorMessage =
+            errorFieldName === 'client_email' && clientEmailLocalPart === 'osd-ccs-admin'
+              ? `The field '${errorFieldName}' must be a Google Cloud service account email ending in '.iam.gserviceaccount.com'.`
```
LiorKGOW, line 1433: Nitpick & non-blocking: 
We know that `errorFieldName` === 'client_email' in this case (also on line 1429) so we could just put it's value directly 😄

LiorKGOW (approved):
I don't want to delay you on this PR, so I approve.
Left a couple of non-blocking comments for you to consider

src/common/validators.ts
DcooperRH: Good point, but I'm to keep this as is just incase the name of the 'client_email' field ever changes this very small part will be more dynamic and we wont have to update it in two places

PR #462, opened 2026-04-20
OCMUI-4397 - HCP: Default autoscaling values when adding a new machine pool should be min=2, max=2, OCMUI-4521 - HCP max_replica should no less than 1 for autoscale nodepool of hcp cluster

DcooperRH:
# Summary

This PR sets the default value for min replicas, max replicas and compute node count to 2 when adding a new machine pool. It also ensures that the max nodes cannot be scaled down to 0.

Assisted by Cursor/claude-4.6-opus

# Jira

Fixes [OCMUI-4397](https://issues.redhat.com/browse/OCMUI-4397) 
Fixes [OCMUI-4521](https://issues.redhat.com/browse/OCMUI-4521) 

# How to Test
**OCMUI-4397**
1. Create a new HCP cluster or go to the cluster details of an existing one
2. Go to the Machine pools tab in the cluster details 
3. Click 'Add Machine pool'
4. Check that the compute node count default is 2
5. Click the enable autoscaling checkbox
6. Check that the min and max replicas default as 2 

**OCMUI-4521**
1. Create a new HCP cluster or go to the cluster details of an existing one
2. Go to the Machine pools tab in the cluster details 
3. Click 'Add Machine pool' or edit an existing machine pool
4. Enable autoscaling and make sure the max nodes field cannot be scaled down to 0

# Screen Captures

| ocmui4397_before | ocmui4397_after |

# Review process

## QE Reviewer

[OCMUI-4397]: https://redhat.atlassian.net/browse/OCMUI-4397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OCMUI-4521]: https://redhat.atlassian.net/browse/OCMUI-4521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/EditMachinePoolModal.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/EditMachinePoolModal.tsx
@@ -380,7 +380,6 @@ const EditMachinePoolModal = ({
                       machinePools={machinePoolsResponse || []}
                       machineTypes={machineTypesResponse}
                       allow249NodesOSDCCSROSA={allow249NodesOSDCCSROSA}
-                      isEdit={isEdit}
                     />
                     <AutoRepairField cluster={cluster} />
                     {imdsSectionFeature && !isEdit && isHypershift ? (
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/EditMachinePoolModal.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/EditMachinePoolModal.tsx
@@ -172,6 +172,8 @@ const EditMachinePoolModal = ({
       );
   }
 
+  const isMaxReached = hcpMaxDifference === 0;
+
   const { initialValues, validationSchema } = useMachinePoolFormik({
     machinePool: currentMachinePool,
     cluster,
@@ -208,17 +210,6 @@ const EditMachinePoolModal = ({
     setIsEdit(getIsEditValue());
   }, [getIsEditValue]);
 
-  // Checks if max nodes amount is reached for add machine pool nodes
-  const isMaxReached =
-    isHypershift &&
-    machinePoolsResponse &&
-    getNodeCount(
-      machinePoolsResponse,
-      isHypershift,
-      currentMachinePool?.id,
-      currentMachinePool?.instance_type,
-    ) === getMaxNodesHCP(cluster.version?.raw_id);
-
   const { mutateAsync: editCreateMachineOrNodePoolMutation } = useEditCreateMachineOrNodePools(
     isHypershift,
     cluster,
@@ -238,6 +229,7 @@ const EditMachinePoolModal = ({
     machineTypesLoading,
     tabKey: 1,
     initialTabContentShown: STARTING_TAB_KEY === 1,
+    isMaxReached,
   });
 
   const [maintenanceTab, maintenanceContent] = useMaintenanceSubTab({
@@ -380,6 +372,7 @@ const EditMachinePoolModal = ({
                       machinePools={machinePoolsResponse || []}
                       machineTypes={machineTypesResponse}
                       allow249NodesOSDCCSROSA={allow249NodesOSDCCSROSA}
+                      isMaxReached={isMaxReached}
                     />
                     <AutoRepairField cluster={cluster} />
                     {imdsSectionFeature && !isEdit && isHypershift ? (
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/EditMachinePoolModal.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/EditMachinePoolModal.tsx
@@ -159,11 +159,25 @@ const EditMachinePoolModal = ({
   const [currentMachinePool, setCurrentMachinePool] = React.useState<MachinePool>();
   const [isEdit, setIsEdit] = React.useState<boolean>(getIsEditValue());
   const [activeTabKey, setActiveTabKey] = React.useState<string | number>(STARTING_TAB_KEY);
+
+  let hcpMaxDifference;
+  if (machinePoolsResponse && isHypershift) {
+    hcpMaxDifference =
+      getMaxNodesHCP(cluster.version?.raw_id) -
+      getNodeCount(
+        machinePoolsResponse,
+        isHypershift,
+        currentMachinePool?.id,
+        currentMachinePool?.instance_type,
+      );
+  }
+
   const { initialValues, validationSchema } = useMachinePoolFormik({
     machinePool: currentMachinePool,
     cluster,
     machinePools: machinePoolsResponse || [],
     machineTypes: machineTypesResponse,
+    hcpMaxDifference,
   });
 
   const isGCP = cluster?.cloud_provider?.id === CloudProviderType.Gcp;
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/EditMachinePoolModal.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/EditMachinePoolModal.tsx
@@ -366,6 +366,7 @@ const EditMachinePoolModal = ({
                       machinePools={machinePoolsResponse || []}
                       machineTypes={machineTypesResponse}
                       allow249NodesOSDCCSROSA={allow249NodesOSDCCSROSA}
+                      isEdit={isEdit}
                     />
                     <AutoRepairField cluster={cluster} />
                     {imdsSectionFeature && !isEdit && isHypershift ? (
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/AutoscaleMaxReplicasField.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/AutoscaleMaxReplicasField.tsx
@@ -54,6 +54,9 @@ const AutoscaleMaxReplicasField = ({
     if (!isEdit && isHypershift && maxNodes < 2) {
       helpers.setValue(maxNodes);
     }
+    if (maxNodes === 0) {
+      helpers.setTouched(true, false);
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/AutoscaleMaxReplicasField.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/AutoscaleMaxReplicasField.tsx
@@ -19,7 +19,6 @@ type AutoscaleMaxReplicasFieldProps = {
   cluster: ClusterFromSubscription;
   maxNodes: number;
   mpAvailZones?: number;
-  isHypershift?: boolean;
 };
 
 const fieldId = 'autoscaleMax';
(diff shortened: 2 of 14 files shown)


jmekkatt (changes requested):
@DcooperRH , If the cluster has already reached the maximum supported node count(500), adding a new machine pool does not trigger a validation error on the compute node count field by default. This can confuse users, as the “Add machine pool” button remains disabled even when all required inputs are provided.
The same issue occurs when autoscaling is enabled. In both scenarios, the default value of 2 contributes to this confusion.
The proposal is to display a validation error on the compute node count, minimum node count, and maximum node count fields to clearly inform the user about the limitation.
See the attached recordings

https://github.com/user-attachments/assets/46067c7a-6310-4929-9350-8ad5befa775e

DcooperRH:
> @DcooperRH , If the cluster has already reached the maximum supported node count(500), adding a new machine pool does not trigger a validation error on the compute node count field by default. This can confuse users, as the “Add machine pool” button remains disabled even when all required inputs are provided. The same issue occurs when autoscaling is enabled. In both scenarios, the default value of 2 contributes to this confusion. The proposal is to display a validation error on the compute node count, minimum node count, and maximum node count fields to clearly inform the user about the limitation. See the attached recordings

Hi @jmekkatt , Good catch. The field should now validate when initially loaded to trigger the validation errors

jmekkatt (approved):
Thanks @DcooperRH for the change. LGTM
Note: Could you please help with the review on test case spec [change](https://github.com/RedHatInsights/uhc-portal/pull/467) that also covers this change in day2? Thank you.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/AutoscaleMaxReplicasField.tsx
```diff
@@ -47,6 +47,11 @@ const AutoscaleMaxReplicasField = ({
     helpers.setTouched(true, false);
   };
 
+  React.useEffect(() => {
```
lizagilman: @DcooperRH IMHO initializing the form with values that don’t pass validation is not great UX. If the default isn’t valid, we should fall back to the minimum valid value instead. Also these useEffect hooks will also run when editing an existing machine pool, making validation errors display immediately.

lizagilman (changes requested):

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/AutoscaleMaxReplicasField.tsx
t-hendricks: I second this.

DcooperRH: I just pushed a commit that if the maxNodes value is < 2, when adding a new hypershift machine pool, the value initializes as the maxNodes count. So if 499 nodes have been created already, the nodes values initialize as 1. If 500 nodes have been created already, the nodes values initialize as 0. Else, the nodes values initialize as 2. 
wdyt @lizagilman @t-hendricks @jmekkatt

t-hendricks (approved):
Tested the new changes locally. LGTM!

jmekkatt (changes requested):
Hi @DcooperRH , I'm revoking my previous approval due to below issue seen with latest revision changes.
The Edit machine pool dialog holds wrong min value when the cluster reached to the max node count quota.
I have a rosa hosted cluster with below machine pools
- first machine pool with compute node count 498
- second and third machine pool with min = 0 and max = 1 with autoscaling enabled.
If user edit second /third machine pool, the dialog set min=1 automatically also the save button enabled. This is wrong behavior.

https://github.com/user-attachments/assets/899b5476-3f2c-4581-bfb0-944518914430

DcooperRH:
> Hi @DcooperRH , I'm revoking my previous approval due to below issue seen with latest revision changes. The Edit machine pool dialog holds wrong min value when the cluster reached to the max node count quota. I have a rosa hosted cluster with below machine pools

Hi @jmekkatt , this should be resolved. The isEdit prop was not being passed correctly

jmekkatt:
Thanks @DcooperRH  for the changes.
One last issue observed during regression:
For a ROSA HCP cluster with the following machine pools:
m1: min = 0, max = 1
m2: min = 0, max = 1
m3: node count = 2

Issue 1:
When editing machine pool m3 and enabling autoscaling, the default values are set to min = 0 and max = 0. Ideally, this should default to 2 (both min and max).
Issue 2:
Similarly, when editing machine pool m1 and disabling autoscaling, the compute node count is set to 0 instead of the expected default value (e.g., 2). Could you please take a look?

DcooperRH:
> Issue 1: When editing machine pool m3 and enabling autoscaling, the default values are set to min = 0 and max = 0. Ideally, this should default to 2 (both min and max). Issue 2: Similarly, when editing machine pool m1 and disabling autoscaling, the compute node count is set to 0 instead of the expected default value (e.g., 2). Could you please take a look?

@jmekkatt I believe this is fixed in my last commit, also @lizagilman @t-hendricks I've added a fix for bug OCMUI-4521 to this PR. Which fixes the issue of max_replicas being allowed to scale down to 0. They should not be able to be scaled below 1

jmekkatt:
> > Issue 1: When editing machine pool m3 and enabling autoscaling, the default values are set to min = 0 and max = 0. Ideally, this should default to 2 (both min and max). Issue 2: Similarly, when editing machine pool m1 and disabling autoscaling, the compute node count is set to 0 instead of the expected default value (e.g., 2). Could you please take a look?
> 
> @jmekkatt I believe this is fixed in my last commit, also @lizagilman @t-hendricks I've added a fix for bug OCMUI-4521 to this PR. Which fixes the issue of max_replicas being allowed to scale down to 0. They should not be able to be scaled below 1

Hi @DcooperRH , I'm a bit confused here. Our primary goal is to align the behavior with the backend/CLI, which is why we allow setting the max replica count to zero (as long as the cluster has a total untainted node count ≥ 2). We can still set the max replica count to zero via the CLI.
Will this restriction also apply to the CLI? If not, it could lead to inconsistent and undesirable behavior in the OCM UI.
cc : @lizagilman

jmekkatt (changes requested):
Hi @DcooperRH,
Currently, when the cluster reaches the maximum node count limit and a user tries to create another machine pool with autoscaling enabled, both the min and max values are set to 0 by default. However, the validation error on the max field shows: "Max nodes must be greater than 0."
This is confusing, as the system sets the default max value to 0 while also flagging it as invalid, without clearly indicating the actual issue.

https://github.com/user-attachments/assets/a2818e91-2aea-40c8-be7f-4fd2805b4437

In this scenario, it would be better to disable the min and max fields and display a clear message indicating that the quota limit has been reached.

jmekkatt:
Hi @DcooperRH , Another scenario related to quota limits when editing a machine pool:
Consider a ROSA hosted cluster with the following machine pools:
Machine pool m1 with compute node count = 1
Machine pool m2 with compute node count = 499
In this case, the total maximum node count per cluster limit has already been reached.
When a user tries to edit machine pool m1 and enables autoscaling, the min and max values are automatically set to 2. However, the max value is invalid.
This behavior seems incorrect. The min and max values should be dynamically calculated based on the remaining quota. In this case, they should default to 1 instead of 2.

https://github.com/user-attachments/assets/4e6fa510-2b8c-4df0-a304-a11e081cf0ef

DcooperRH:
@jmekkatt I believe the above two concerns have been addressed. Max nodes field validation error will read as 'Max nodes limit has been reached'

jmekkatt:
> @jmekkatt I believe the above two concerns have been addressed. Max nodes field validation error will read as 'Max nodes limit has been reached'

Hi @DcooperRH 
There are still inconsistencies in how values are handled when the cluster has reached its maximum node count limit. In a ROSA HCP cluster with a total node limit of 500 already reached, adding a new machine pool with autoscaling enabled currently defaults to min = 0 and max = 0, along with a validation error stating that the max node limit has been reached. However, since max = 0 is no longer supported in HCP and must be ≥ 1, the expected behavior is to default max to 1 and show the same validation error.
Similarly, when editing an existing machine pool and enabling autoscaling, the current behavior sets min = 2 and max = 2 with a validation error. Instead, these values should be dynamically derived based on remaining capacity—min should default to 0 and max to the minimum supported value of 1, along with the “max nodes limit has been reached” error.

https://github.com/user-attachments/assets/aee91e67-2175-4dae-836c-252e2e3f0ac8

lizagilman (approved):

lizagilman:
Current version of the code LGTM. Dylan if/when you push more changes to address Jaya's concern I"ll take another look.

DcooperRH:
> Hi @DcooperRH There are still inconsistencies in how values are handled when the cluster has reached its maximum node count limit. In a ROSA HCP cluster with a total node limit of 500 already reached, adding a new machine pool with autoscaling enabled currently defaults to min = 0 and max = 0, along with a validation error stating that the max node limit has been reached. However, since max = 0 is no longer supported in HCP and must be ≥ 1, the expected behavior is to default max to 1 and show the same validation error. Similarly, when editing an existing machine pool and enabling autoscaling, the current behavior sets min = 2 and max = 2 with a validation error. Instead, these values should be dynamically derived based on remaining capacity—min should default to 0 and max to the minimum supported value of 1, along with the “max nodes limit has been reached” error.
> 


Hi @jmekkatt , thanks for catching this. I believe all issues should be addressed now. @lizagilman

jmekkatt (approved):
Thanks for all the efforts @DcooperRH , Changes lgtm.
Note:  I have seen multiple improvements regarding the error validation in different circumstances on edit/add machine pool models. I would create seperate ticket for the same as they are not in current scope of this change. Thank you.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/AutoscaleMaxReplicasField.tsx
lizagilman: I still think we should probably avoid this `useEffect`  and the `  // eslint-disable-next-line react-hooks/exhaustive-deps` hack ..  still not great ux (in the case where nodes values initialize to 0 --> resulting errors appear without user interaction), and also it's against code guidelines, given there are other solutions. 
I can suggest -
1. disable the autosacle checkbox and a tooltip explaining max nodes reached. Since this is an edge case I think it doesn't worth increasing code complexity and descresing code hygiene with a workaround. 
or
2. Use `validateOnMount` combined with `initialTouched` props on the `<<Formik<EditMachinePoolValues> .. >` to avoid the `useEffect` .

IMHO option 1 is preferred.
@DcooperRH Can you give it another shot?

Edit: I happened to have a chat with a UX engineer, he suggests it's better to disable the checkbox entirely and instead of tooltip show an inline alert explaining max nodes reached. He says inline alert is prefred over a tooltip because user may not realize he needs to hover over the checbox, and with inline alert the reason is shown right away. Makes sense to me, but a tooltip can also work.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/EditMachinePoolModal.tsx
lizagilman:
```suggestion
// Checks if max nodes amount is reached for add machine pool nodes
const isMaxReached = hcpMaxDifference === 0;
```

This is not a nitpick 🙏  we should avoid duplicate calculations. We already calculate max nodes and max nodes related logic in many different places across the app (it's a big tech debt 🤦‍♀️ ) 
Let's try to keep it minimal with new code.

lizagilman (changes requested):

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/components/EditMachinePoolModal/fields/AutoscaleMaxReplicasField.tsx
DcooperRH: Ok I've added the Alert shown below for when the field is disabled. I've also removed the logic behind the useEffect triggering the max nodes validation, which was also removed from the Yup validation scheme as the 'max nodes has been reached' validation is now unreachable. @jmekkatt sorry but I think you will have to re-review this pr due to this addition  
Screenshot 2026-05-06 at 1 48 41 PM

lizagilman (approved):

jmekkatt (approved):
Re-review has done. LGTM.

PR #471, opened 2026-04-22
OCMUI-4319: Add exclude namesapce selector to OSD GCP wizard (Day 1)

zherman0:
# Summary

For the OSD GCP wizard, under custom application ingress in networking, added Exclude namespace selectors using a generalized FormKeyValueList (custom column labels, add button, valueAriaLabel, validators). Gated with ocmui-gcp-exclude-namespace-selectors and shown only for GCP.

Analytics: OcmIngressExcludeNamespaceSelectorsSet fires after a successful create when selectors are configured, via useTrackOcmIngressExcludeNamespaceSelectorsOnClusterCreate in CreateOsdWizard

Updated Shared components: FormKeyLabelList, FormKeyLabelKey, and FormKeyLabelValue gained optional valueAriaLabel; errors use FormGroupHelperText for consistent PF error + icon behavior. With these updates, now the component is generic and can be reused anywhere.

# Jira

Fixes [OCMUI-4319](https://issues.redhat.com/browse/OCMUI-4319)

# Additional information

`parseFormExcludeNamespaceSelectorsToApi` maps form rows → excluded_namespace_selectors on the default ingress item in submitOSDRequest when non-empty.

Added lots and lots of test as most of this did not have coverage.

# How to Test
0. Make sure the feature flag is turned on: `ocmui-gcp-exclude-namespace-selectors`
1. Go to the OSD wizard, select annual or GCP marketplace with CCS selected.
2. Select GCP (if needed), select WIF (although SA will work it does not matter), pick a WIF config or make one 
3. Fill out cluster details, use what ever value you want
4.  Use default values until you get to the Network  configuration step. Scroll down to "Application ingress settings" and selected "Custom Settings"
5.  You will now see "Exclude namespace selectors" (not to be confused with "Excluded namespaces")
6. Fill in some values - it does not matter what you enter as long as they have the correct format (end with letter/number, no special chars)
7. Example values would be: 
```
{ "key": "department", "values": ["finance", "HR"] },
    { "key": "type", "values": ["customer"] }
```
8. If you enter `openshift-console` or `openshift-authentication`, you should see an error as those values are protected
9. Continue on in the wizard with default values until you get to the review page.  There will be a new line item in the networking section called "Exclude namespace selectors"
10. Submit your cluster and let it build
11. Now, since there is no day 2 yet, you will have to use the command line to verify the values are added:
```
ocm get "/api/clusters_mgmt/v1/clusters/$(ocm describe cluster mycluster --json | jq -r '.id')/ingresses"
```
or if you want the separate steps because you are cool:
```
ocm describe cluster mycluster --json | jq -r '.id'
ocm get /api/clusters_mgmt/v1/clusters/{id}/ingresses

```

# Screen Captures
New Input:

Screenshot From 2026-04-22 11-34-40

New input with Error:

Screenshot From 2026-04-22 11-35-10

Review page:

Screenshot From 2026-04-24 17-38-39

# Review process

## QE Reviewer

[OCMUI-4319]: https://redhat.atlassian.net/browse/OCMUI-4319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dtaylor113 (changes requested):
Hi @zherman0, idk, on the Review step, I was really expecting the selectors to be in that 'rounded blue pill' that we currently use for Node Labels.  However, I see that existing "Excluded namespaces" is just plain text as well.

However, IMO the problem is worse expressing selectors using pure text, evertyhing seems to run on together: "dept: finance, HR, IT; prod: HCC, FEC, FLUFFY-SHEET; asdfasdf: openshift-consoles" -hard to parse visually.   If each selector was in it's own 'pill' it would, IMO, be easy to read.  If we change the exclude selectors, we should also change the existing display of excluded namespaces as well.  wdyt?

Image

dtaylor113:
Verified the "How to Test" steps and see the updated items in the OCM cmd.
 Still need to review the code.
 ```
"items": [
    {
      "kind":"Ingress",
      "href":"/api/clusters_mgmt/v1/clusters/2psn14b6cfa4dotvflgje0pq9d0m89s3/ingresses/p4b4",
      "id":"p4b4",
      "listening":"external",
      "default":true,
      "dns_name":"apps.dt-osd-2.wgvm.s2.devshift.org",
      "excluded_namespaces": [
        "foo",
        "bar",
        "this",
        "that"
      ],
      "excluded_namespace_selectors": [
        {
          "key":"dept",
          "values": [
            "finance",
            "HR",
            "IT"
          ]
        },
        {
          "key":"prod",
          "values": [
            "HCC",
            "FEC",
            "FLUFFY-SHEET"
          ]
        },

```

File: src/common/validators.ts [hunk 1]
```diff
@@ -1726,6 +1715,104 @@ const validateLabelKey = (
...
+    }
+  }
+
+  for (let i = 0; i < parts.length; i += 1) {
+    const part = parts[i];
+    if (
+      part != null &&
+      PROTECTED_DEFAULT_ROUTER_NAMESPACE_SELECTOR_VALUES.has(part.toLowerCase())
+    ) {
+      return 'Do not exclude openshift-console or openshift-authentication namespaces; they are vital to cluster operations.';
```
DcooperRH, line 1819: non-blocking, but the usage of 'vital' here seems weird. What about 'they are necessary for cluster operations' or 'they are required for cluster operations'

DcooperRH:
The validation font here seems pretty large as compared to the Node Labels validation 

**Node Labels**
Screenshot 2026-04-24 at 4 09 55 PM


**Exclude namespace selectors**
Screenshot 2026-04-24 at 4 10 33 PM

src/common/validators.ts
zherman0: @dtaylor113 @DcooperRH - That is the exact quote from the ticket, but I have no issue changing the wording. Dave, thoughts?

zherman0:
> The validation font here seems pretty large as compared to the Node Labels validation
> 

@DcooperRH - I rewrote the common component FormKeyLabelValue to use the right helper, but I did not realize that the FormKeyLabelKey component had the same issue!  I have fixed it _and_ I added a story book so it is easier to test.  The story book is all AI but it  looks good to me.  _yarn storybook_ and then look at FormKeyValueList
Screenshot From 2026-04-24 17-19-29

zherman0:
> Hi @zherman0, idk, on the Review step, I was really expecting the selectors to be in that 'rounded blue pill' that we currently use for Node Labels. However, I see that existing "Excluded namespaces" is just plain text as well.

@dtaylor113 - I love it!  I updated both excluded namespaces and excluded namespace selectors.  I also made a storybook for the wizard review page.  I would hope it is okay to check this in as it makes it super easy to mmake changes to the review page on the fly.  _yarn storybook_ 

Update:  BAH!!!! Now I have to do route selectors as well! Done.

Screenshot From 2026-04-24 17-38-39

src/common/helpers.ts
```diff
@@ -45,10 +45,11 @@ const asArray = <T>(itemOrList: T | T[]): T[] => {
 };
 
 const stringToArray = (str?: string) => str && str.trim().split(',');
-const stringToArrayTrimmed = (str: string) =>
-  asArray(stringToArray(str))
-    .map((ns) => ns?.trim())
-    .filter((ns) => !!ns);
+const stringToArrayTrimmed = (str: string): string[] =>
```
zherman0, line 48: code rabbit aasked for me to change this since it takes in a string.  Made sense to me so I tweaked it.

src/components/clusters/wizards/common/ReviewCluster/reviewValues.jsx
```diff
@@ -383,13 +382,73 @@ const reviewValues = {
   },
   defaultRouterSelectors: {
     title: 'Route selectors',
-    valueTransform: (value) =>
-      value ? routeSelectorsAsString(strToKeyValueObject(value, '')) : 'None specified',
+    valueTransform: (value) => {
```
zherman0, line 385: This was a minimal change  but it was suggested to show everything as blue labels like nodeLabels does.

src/components/clusters/wizards/common/ReviewCluster/reviewValues.stories.tsx
```diff
@@ -0,0 +1,252 @@
+import React from 'react';
```
zherman0, line 1: So, testing visual changes on the last page of the wizard is maybe the worst part of our job.  This story should make that 10x easier.  You are welcome.

zherman0:
@dtaylor113 @DcooperRH -  I added another 500 lines to this story and a few more files - sorry about that, but they are all for storybook which is such a cool tool.  Feel free to try it out for the FormKeyValueList component and now the reviewValues screen on the wizard (so helpful).

src/queries/featureGates/featureConstants.ts
```diff
@@ -46,6 +46,7 @@ export const OSD_FOR_GOOGLE_CLOUD = 'ocmui-osd-wizard-for-google-cloud';
 export const FIPS_FOR_HYPERSHIFT = 'ocmui-fips-for-hypershift';
 export const Y_STREAM_CHANNEL = 'ocmui-y-stream-channel';
 export const GCP_DNS_ZONE = 'ocmui-gcp-dns-zone';
+export const EXCLUDE_NAMESPACE_SELECTORS = 'ocmui-exclude-namespace-selectors';
```
LiorKGOW: Hey Zac 👋🏼
As I am working on the Day2 part of this feature, I put myself as reviewer for this one so we'll be aligned on the shared resources related to this feature. I hope it is ok with you 😄 

I have a small suggestion to append to the name the feature flag: 
**GCP**_EXCLUDE_NAMESPACE_SELECTORS = ocmui-**gcp-** exclude-namespace-selectors

As I also see it for `GCP_DNS_ZONE = 'ocmui-gcp-dns-zone'`, I think we should specify this explicitly on the feature flag name & value to make this clear both from Unleash and from the code perspectives, that this feature **revolves GCP clusters only**. We could also consider appending a `osd-` too, but as we only have osd gcp clusters, I don't think it would be necessary 😇 

Let me know what you think about my suggestion @zherman0 

```suggestion
export const GCP_EXCLUDE_NAMESPACE_SELECTORS = 'ocmui-gcp-exclude-namespace-selectors';
```

zherman0: Done

src/components/clusters/wizards/osd/Networking/DefaultIngressFields.tsx
```diff
@@ -24,7 +34,9 @@ type DefaultIngressFieldsProps = {};
 
 /* So far used in Day 1 flow only (contrary to the version 1 CreateOSDWizard */
 export const DefaultIngressFields: React.FC<DefaultIngressFieldsProps> = () => {
-  const { setFieldValue, getFieldMeta, getFieldProps } = useFormState();
+  const { setFieldValue, getFieldMeta, getFieldProps, values } = useFormState();
+  const isExcludeNamespaceSelectorsEnabled = useFeatureGate(GCP_EXCLUDE_NAMESPACE_SELECTORS);
+  const isGcp = values[FieldId.CloudProvider] === CloudProviderType.Gcp;
```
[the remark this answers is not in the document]
zherman0: Edge case but also valid.  Changed code to handle.

src/common/validators.ts
dtaylor113: sure, that seems fine.

src/common/validators.test.ts
```diff
@@ -79,4 +82,165 @@ describe('validators', () => {
...
+    });
+
+    it('returns an error when the key is not a valid Kubernetes label key', () => {
+      const message = validateExcludeNamespaceSelectorKey(
+        'Invalid_Label_Key!',
+        { [field]: [{ key: 'Invalid_Label_Key!', value: 'v' }] },
+        undefined,
+        keyName(0),
+      );
+      expect(message).toBeDefined();
```
dtaylor113: Should we test the message contents?

src/common/validators.test.ts
```diff
@@ -79,4 +82,165 @@ describe('validators', () => {
...
+      expect(message).not.toBe('');
+    });
+
+    it('returns undefined for a valid unique key', () => {
```
dtaylor113, line 114: This seems incorrect?  returns undefined for a valid key?

src/common/validators.ts
DcooperRH: Oh I must've overlooked that on the ticket

src/common/validators.test.ts
```diff
@@ -79,4 +82,165 @@ describe('validators', () => {
...
+        validateExcludeNamespaceSelectorValue(
+          '   ',
+          { [field]: [{ key: 'env', value: '' }] },
+          undefined,
+          valueName(0),
+        ),
+      ).toBe('Enter at least one value, separated by commas.');
+    });
+
+    it('returns undefined for comma-separated label values', () => {
```
dtaylor113, line 240: This also doesn't seem correct "returns undefined for comma-separated label values"

src/components/clusters/ClusterDetailsMultiRegion/components/Networking/components/ApplicationIngressCard/ExcludeNamespaceSelectorsPopover.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+import {
+  ExcludeNamespaceSelectorsHelpText,
+  ExcludeNamespaceSelectorsPopover,
+} from './ExcludeNamespaceSelectorsPopover';
+
+describe('ExcludeNamespaceSelectorsPopover', () => {
+  describe('ExcludeNamespaceSelectorsHelpText', () => {
+    it('matches the inline helper copy used under the field', () => {
+      expect(ExcludeNamespaceSelectorsHelpText).toBe(
+        'Namespaces matching any of these label selectors will be excluded from the default ingress controller.',
```
dtaylor113, line 14: This seems like an incredibly useless and potentially breaking if message changes, but if needed for code coverage...

src/components/clusters/ClusterDetailsMultiRegion/components/Networking/components/ApplicationIngressCard/ExcludeNamespaceSelectorsPopover.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import PopoverHint from '~/components/common/PopoverHint';
+
+export const ExcludeNamespaceSelectorsHelpText =
+  'Namespaces matching any of these label selectors will be excluded from the default ingress controller.';
+
+export const ExcludeNamespaceSelectorsPopover = () => (
+  <PopoverHint
+    title="Exclude namespace selectors"
+    maxWidth="30rem"
```
dtaylor113: This seems like an anti-pf pattern.  Is there a PF default enum like LG, MED, SM etc.. or another prop. on PopoverHint to set defined standard sizes?

src/components/clusters/wizards/common/ReviewCluster/reviewValues.stories.tsx
dtaylor113: Thanks, that storybook story does make review easy, quick, and 'complete' ie showing all the fields without requiring specific configurations.

src/components/clusters/wizards/common/ReviewCluster/reviewValues.test.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+    it('shows None specified when parsing yields no pairs', () => {
+      expect(valueTransform('   ')).toBe('None specified');
+    });
+
+    it('renders each route selector as a blue label', () => {
+      render(<>{valueTransform('route=external,shard=blue')}</>);
+      expect(screen.getByText('route = external')).toBeInTheDocument();
+      expect(screen.getByText('shard = blue')).toBeInTheDocument();
+    });
```
dtaylor113, line 28: What about a test for "'route=external,shard='"?

src/components/clusters/wizards/common/ReviewCluster/reviewValues.test.jsx
```diff
@@ -0,0 +1,96 @@
...
+      expect(valueTransform([])).toBe('None specified');
+    });
+
+    it('shows None specified when only placeholder or whitespace keys are present', () => {
+      expect(
+        valueTransform([
+          { id: '1', key: '', value: '' },
+          { key: '   ', value: 'x' },
+        ]),
+      ).toBe('None specified');
```
dtaylor113, line 55: Hmm...I'm not crazy about the valueTransform() function return "None specified" for "Invalid data", seems like it obfuscating something.  I realize form field validation should catch this before the review page, but...

src/components/clusters/wizards/common/ReviewCluster/reviewValues.test.jsx
```diff
@@ -0,0 +1,96 @@
...
+      render(
+        <>
+          {valueTransform([
+            { key: 'department', value: 'finance' },
+            { key: 'type', value: 'customer' },
+          ])}
+        </>,
+      );
+      expect(screen.getByText('department = finance')).toBeInTheDocument();
+      expect(screen.getByText('type = customer')).toBeInTheDocument();
```
dtaylor113, line 83: How about a test for:
{valueTransform([
  { key: 'department', value: 'finance' },
   { key: '', value: 'a' },
])}

src/components/clusters/wizards/common/ReviewCluster/reviewValues.test.jsx
```diff
@@ -0,0 +1,96 @@
...
+    });
+  });
+
+  describe('defaultRouterExcludedNamespacesFlag', () => {
+    const { valueTransform } = reviewValues.defaultRouterExcludedNamespacesFlag;
```
dtaylor113, line 88: What is "defaultRouterExcludedNamespacesFlag"?
Seems like a repeat of other unit tests.

src/components/common/FormikFormComponents/FormKeyValueList/FormKeyValueList.tsx
```diff
@@ -17,58 +18,106 @@ import FormKeyLabelValue from './FormKeyLabelValue';
...
+  validateValue = validateLabelValue,
+  addButtonDisabledTooltip = DEFAULT_ADD_BUTTON_DISABLED_TOOLTIP,
+}: FormKeyValueListProps) => {
+  const { values, setFieldValue, setFieldTouched, getFieldProps, getFieldMeta, validateForm } =
+    useFormState();
+
+  const fieldRows = (values as FormikValues)[arrayFieldName] as KeyValueRow[];
+  const rows = Array.isArray(fieldRows) ? fieldRows : [];
+
+  const hasInvalidKeys = (fieldsArray: KeyValueRow[]) =>
```
dtaylor113, line 86: This seems like a significant refactor of src/components/common/FormikFormComponents/FormKeyValueList/FormKeyValueList.tsx, what other components use this FormKeyValueList.tsx, we should test it still works for other usages.

src/common/validators.test.ts
zherman0: It is a long message, but if you think that is better.  I worry about it being brittle, but I changed it anyway.

DcooperRH (approved):
Went through and retested with the new changes and didn't find any errors or ux concerns. Code LGTM

src/common/validators.test.ts
zherman0: That is how the error helpers work - undefined means no error, while we display the message that gets defined.  The prior test case shows this perfectly.

src/common/validators.test.ts
zherman0: It is correct for validation

src/components/clusters/ClusterDetailsMultiRegion/components/Networking/components/ApplicationIngressCard/ExcludeNamespaceSelectorsPopover.test.tsx
zherman0: Not sure what to say on this.  I need to cover that the message is being tested.

src/components/clusters/ClusterDetailsMultiRegion/components/Networking/components/ApplicationIngressCard/ExcludeNamespaceSelectorsPopover.tsx
zherman0: Agreed.  Removing now.

zherman0: Actually, I see I copied it from the Excluded namespaces popover  and the Router selector popover bith done 22months ago.  If I change it, they won't all be the same.

src/components/clusters/wizards/common/ReviewCluster/reviewValues.test.jsx
zherman0: ok, I did add all the unit tests, but I did not write the router stuff so I was not trying to add more work to this PR.  I will add the test but  I am not sure it is my responsibility to write every possible test for other features.  
UPDATE: I added the test and it does pass.

src/components/clusters/wizards/common/ReviewCluster/reviewValues.test.jsx
zherman0: Not sure what you would like me to do here.

src/components/clusters/wizards/common/ReviewCluster/reviewValues.test.jsx
zherman0: That would throw an error before we get here so this test does not make sense.

src/components/clusters/wizards/common/ReviewCluster/reviewValues.test.jsx
zherman0: I am just validating the flag is enabled.  It is used in a lot of places for the excluded namespace for the router.  Again, there were no tests here and I was just checking the review values, not any other logic that should be handled before this page.

src/common/validators.test.ts
dtaylor113: got it, undefined === ok 👍

src/common/validators.test.ts
dtaylor113: I understand now undefined means "validation ok"

src/components/clusters/ClusterDetailsMultiRegion/components/Networking/components/ApplicationIngressCard/ExcludeNamespaceSelectorsPopover.test.tsx
dtaylor113: Yeah, I think we need a post-scrum about making unit tests simply for codecov and also why codecov is a requirement Check which needs to pass(?)

src/components/clusters/ClusterDetailsMultiRegion/components/Networking/components/ApplicationIngressCard/ExcludeNamespaceSelectorsPopover.tsx
dtaylor113: Can you change so that they both use a shared const?

src/components/clusters/wizards/common/ReviewCluster/reviewValues.test.jsx
dtaylor113: ah, sorry, thanks!

src/components/clusters/wizards/common/ReviewCluster/reviewValues.test.jsx
dtaylor113: Ok, I now understand that valueTransform() is just for exclude namespace selector and that field validation in earlier step ensures that this should never be shown.

src/components/common/FormikFormComponents/FormKeyValueList/FormKeyValueList.tsx
```diff
@@ -134,9 +183,9 @@ const FormKeyValueList = ({ push, remove }: ArrayHelpers) => {
           variant="link"
           isInline
           className="formKeyValueList-addBtn"
-          disableReason={hasInvalidKeys(nodeLabels) && nodeKeyValueTooltipText}
+          disableReason={hasInvalidKeys(rows) && addButtonDisabledTooltip}
         >
-          Add additional label
+          {addButtonLabel}
```
dtaylor113, line 244: Hi @zherman0, again, why such a large refactor for keyvaluelist due to exclude namespace selector?

src/components/clusters/ClusterDetailsMultiRegion/components/Networking/components/ApplicationIngressCard/ExcludeNamespaceSelectorsPopover.tsx
zherman0: There are 9 popover files that all have 30rem.  I can change them super easy, but it will add 9 files and 9 areas of the code to test for this PR

src/components/common/FormikFormComponents/FormKeyValueList/FormKeyValueList.tsx
zherman0: In all fairness, it seemed like a good idea to use a common component, but the component was not designed to be used other places.  The initial changes were quite small and straight forward, but things just grew as more functionality and flexibility was needed.  Add in some code rabbit reviews which added even more guardrail checking and this has become much more than a simple refactor.  But, it will be usable for namespace selectors, nodes, and day 2 stuff, plus whatever might come.  There are other places in the code that have built this in place, and I did not want to add yet another version.

src/components/common/FormikFormComponents/FormKeyValueList/FormKeyValueList.tsx
zherman0: This original version was only built for node labels.  These changes like the addButtonLabel give complete control over the component.

src/components/common/FormikFormComponents/FormKeyValueList/FormKeyValueList.tsx
zherman0: I wanted to say more more thing about this - this component is in common/FormikFormCompnents.  It should have been written generically initially or it should have been moved to a node specific directory since all it handled was node labels.

src/queries/featureGates/featureConstants.ts
LiorKGOW: Thanks Zac!

src/components/clusters/ClusterDetailsMultiRegion/components/Networking/components/ApplicationIngressCard/ExcludeNamespaceSelectorsPopover.tsx
dtaylor113: Hi @zherman0, I think this needs to be done as we don't want hardcoded values, and a single const would help with consistency and maintainability.
Let's see how much it adds to CodeCov and may be it's time to loosen up the codecov PR requirements. -thanks

zherman0: @dtaylor113 - Opened https://redhat.atlassian.net/browse/OCMUI-4520 to track this work.  I will remove the maxWidth from this PR and we will address the rest in this follow up ticket.

dtaylor113: Ok, sounds good, thanks

dtaylor113 (approved):
Was able to create a osd gcp cluster with exclude namespace selectors, verified via CLI.
All my review comments addressed.
Code LGTM

LiorKGOW:
Hey @zherman0 
I think we should update the top comment to match [the new FF name & value](https://github.com/RedHatInsights/uhc-portal/pull/471/#discussion_r3147579521)

jmekkatt:
Hi @zherman0 , The current validation behavior for value fields, where the error is shown only on clicking "Next", is not very user-friendly. Users should receive immediate feedback when they leave the value field (on blur), especially since at least one value is required for a selector.
To improve usability:
Show an inline validation error when the value field is left empty after user interaction (on blur).
Additionally,
Disable the “Add selector” action until the current selector has a valid key and at least one value.
Optionally, disable “Next” until all selectors are valid.

https://github.com/user-attachments/assets/d70e0e91-6924-497e-9f01-9437f4a3a8a7

shivanthi-amara (changes requested):
@zherman0 Tested the code changes as per this slack [discussion](https://redhat-internal.slack.com/archives/C02HPLPJRUK/p1777562514239709). Below are my observations. Please take a look.

1) **Form values are not refreshed when the user switches the authentication types**

When the user encounters validations errors for the key and value fields within the "Exclude namespace selectors" section, and the user switches the authentication type, the validation errors for the attributes are not cleared. Please refer to this attached recording.


https://github.com/user-attachments/assets/adddd2da-6189-461c-94b8-275f85ea82e4


2) **AC wording vs actual UI model**

The acceptance criteria lists A warning message is displayed below the field: "Warning: Do not exclude openshift-console or openshift-authentication namespaces as they are vital to cluster operations." 

Actual Result: Code treats openshift-console / openshift-authentication as invalid selector values with a specific error message (see DefaultIngressFields.test.tsx / validators.ts). AC wording uses “Warning:” There is a difference in the implementation details, so it would be good to clarify what is expected here. (reference image attached)

Screenshot 2026-04-30 at 2 30 49 PM


3) **Consistency with “Excluded namespaces”**

Excluded namespaces” has rich helper text under the box. “Exclude namespace selectors” only has the label + popover—no inline helper line under the control.  Also “Exclude namespace selectors” adds keys, multiple rows, duplicates risk, and comma semantics inside the Values column. A pattern that scales down guidance as complexity goes up trains users to guess—or to rely on trial-and-error errors.

Primary hints in visible helper text are typically always available and easier to associate with the control. Popover body content is often hidden until invoked, so it would be a good to a helper text for this field as well. (reference image attached)

Screenshot 2026-04-30 at 2 28 23 PM

zherman0:
> @zherman0 Tested the code changes as per this slack [discussion](https://redhat-internal.slack.com/archives/C02HPLPJRUK/p1777562514239709). Below are my observations. Please take a look.
> 
>     1. **Form values are not refreshed when the user switches the authentication types**

@shivanthi-amara - I don't believe that error should clear.  I have run the same test with Excluded namespaces and it behaves the same way.

https://github.com/user-attachments/assets/1d55656e-55da-4355-a124-e1cd8adf10aa


> 
>     2. **AC wording vs actual UI model**
> 

This was discussed above by Dave and I.  I forgot to update the ticket.  We now only show an error instead of always showing a warning message.



>     3. **Consistency with “Excluded namespaces”**
> 

Excellent point, I will update this immediately.
Screenshot From 2026-04-30 16-05-32

shivanthi-amara:
> > @zherman0 Tested the code changes as per this slack [discussion](https://redhat-internal.slack.com/archives/C02HPLPJRUK/p1777562514239709). Below are my observations. Please take a look.
> > ```
> > 1. **Form values are not refreshed when the user switches the authentication types**
> > ```
> 
> @shivanthi-amara - I don't believe that error should clear. I have run the same test with Excluded namespaces and it behaves the same way.
> 
>  Screencast.From.2026-04-30.15-44-09.mp4 
> > ```
> > 2. **AC wording vs actual UI model**
> > ```
> 
> This was discussed above by Dave and I. I forgot to update the ticket. We now only show an error instead of always showing a warning message.
> 
> > ```
> > 3. **Consistency with “Excluded namespaces”**
> > ```
> 
> Excellent point, I will update this immediately. Screenshot From 2026-04-30 16-05-32

@zherman0  Regarding #1. I think this is an issue as the user switched the authentication type and is navigating through the form fields for the first time. Validation errors follow the field values and in this case, the user did not input the fields, so showing the previous errors might mislead the user.

Btw, did you also address Jaya's [comment](https://github.com/RedHatInsights/uhc-portal/pull/471#issuecomment-4353411868) in your latest fix?

zherman0:
> Regarding https://github.com/RedHatInsights/uhc-portal/pull/1. I think this is an issue as the user switched the authentication type and is navigating through the form fields for the first time. Validation errors follow the field values and in this case, the user did not input the fields, so showing the previous errors might mislead the user.

>Btw, did you also address Jaya's https://github.com/RedHatInsights/uhc-portal/pull/471#issuecomment-4353411868 in your latest fix?

@shivanthi-amara - But my behavior follows the rest of the custom settings as shown in the video, the values stay even though the auth type changes.  Right?  Are you saying that the other custom settings are wrong and we need to change them as well?

As for Jaya's comment, I missed it and I am implementing a fix now.

shivanthi-amara:
> > Regarding #1. I think this is an issue as the user switched the authentication type and is navigating through the form fields for the first time. Validation errors follow the field values and in this case, the user did not input the fields, so showing the previous errors might mislead the user.
> 
> > Btw, did you also address Jaya's [#471 (comment)](https://github.com/RedHatInsights/uhc-portal/pull/471#issuecomment-4353411868) in your latest fix?
> 
> @shivanthi-amara - But my behavior follows the rest of the custom settings as shown in the video, the values stay even though the auth type changes. Right? Are you saying that the other custom settings are wrong and we need to change them as well?
> 
> As for Jaya's comment, I missed it and I am implementing a fix now.

@zherman0 I agree. This appears to be an existing bug affecting the custom settings fields, as the form doesn't reset when switching between authentication types. Since this issue impacts your code changes, I believe it is worthy of a fix. If the scope of the necessary changes is too large for the current PR, we should consider addressing part of it in a new task. What are your thoughts on how to proceed?

zherman0:
> > @shivanthi-amara - But my behavior follows the rest of the custom settings as shown in the video, the values stay even though the auth type changes. Right? Are you saying that the other custom settings are wrong and we need to change them as well?
> > As for Jaya's comment, I missed it and I am implementing a fix now.
> 
> @zherman0 I agree. This appears to be an existing bug affecting the custom settings fields, as the form doesn't reset when switching between authentication types. Since this issue impacts your code changes, I believe it is worthy of a fix. If the scope of the necessary changes is too large for the current PR, we should consider addressing part of it in a new task. What are your thoughts on how to proceed?

@shivanthi-amara - I absolutely prefer to fix it and the others right now, but I am not sure it is the correct behavior.  Why would changing the auth type affect ingress settings - it wouldn't in my mind.  I am worried someone might configure a lot of custom settings and then realize they picked the wrong auth type and then go and change it to only lose their settings.  The custom settings should be completely agnostic to the auth type.  Namespaces are specific to the openshift cluster, not an authentication.

Looking at the code the only thing we "clear" when a different auth type is chosen is the AcknowledgePrereq checkbox.  

I think the correct thing is to not clear them, but the code change would be easy in the GcpByocFields component.  But just because it easy, I still don't think we should do it.

zherman0:
@shivanthi-amara @jmekkatt - Regarding the error validation, I have made the error validation much better but I decided to not report an issue on the first line of values  if they enter a key and then leave the value blank.  That will only trigger on the next press, but they cannot add another line of key/value without filling in the value.  Please give it a try.  The design I have implemented is very intentional at this point, but there might be some debate.  Once the form has been touched, then I will immediately report that a value has to be entered.

Update: I did demo this to UX and we had discussion.  Waiting for the error to show on clicking of the next is how other wizard pages work (i.e. Cluster name on the details page). This is a tricky situation, but they were fine with my updated solution.

https://github.com/user-attachments/assets/1e8389e3-bbfd-4324-a12f-872725532264

shivanthi-amara:
> > > @shivanthi-amara - But my behavior follows the rest of the custom settings as shown in the video, the values stay even though the auth type changes. Right? Are you saying that the other custom settings are wrong and we need to change them as well?
> > > As for Jaya's comment, I missed it and I am implementing a fix now.
> > 
> > 
> > @zherman0 I agree. This appears to be an existing bug affecting the custom settings fields, as the form doesn't reset when switching between authentication types. Since this issue impacts your code changes, I believe it is worthy of a fix. If the scope of the necessary changes is too large for the current PR, we should consider addressing part of it in a new task. What are your thoughts on how to proceed?
> 
> @shivanthi-amara - I absolutely prefer to fix it and the others right now, but I am not sure it is the correct behavior. Why would changing the auth type affect ingress settings - it wouldn't in my mind. I am worried someone might configure a lot of custom settings and then realize they picked the wrong auth type and then go and change it to only lose their settings. The custom settings should be completely agnostic to the auth type. Namespaces are specific to the openshift cluster, not an authentication.
> 
> Looking at the code the only thing we "clear" when a different auth type is chosen is the AcknowledgePrereq checkbox.
> 
> I think the correct thing is to not clear them, but the code change would be easy in the GcpByocFields component. But just because it easy, I still don't think we should do it.

@zherman0 I believe there is a misunderstanding for #1. I presented a scenario where the form's validation state should be updated upon a change event.

We can illustrate this with a simpler case observed elsewhere in the wizard, specifically in the Cluster Update strategy section. When toggling the radio buttons, the previously made selections are cleared. For example, if you set a date and time under the "Recurring updates" radio button and then select the "Individual updates" button, the values you set are cleared when you switch back to "Recurring updates."


Screenshot 2026-05-01 at 10 04 58 AM


**Similarly, the previously entered values and errors for Exclude namespace selectors should be cleared when toggled between the default and custom ingress settings radio buttons.**


Regarding the error validation scenario Jaya shared, I like the approach of disabling the "Add selector" link until the current selector has a valid key and value, but there seems to no validation error shown for the **Values field** if the user enters a improper input in both Key and Values field.

Screenshot 2026-05-01 at 10 27 17 AM

Attached a recording for reference.


https://github.com/user-attachments/assets/71da09df-90c2-41af-95ca-286ef6a820ae

zherman0:
@shivanthi-amara - I have address your comments so now we clear the values in the custom ingress settings (all of them, not just NS selectors).  I also updated the validation rules to catch the errors that you indicated.  The code actually automatically removed the white space and extra commas, but I think that is confusing and let the user think those were valid entries. Thanks

shivanthi-amara:
@zherman0 Verified the changes based on the [testcase](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-88885) and so far everything looks good to me except the UI and CLI validation error rules that we discussed. Conversation [link](https://redhat-internal.slack.com/archives/C0AEPA2CC3U/p1777670416617939).

@jmekkatt Just an FYI,  kindly review the changes and let me know your feedback.

jmekkatt:
> Hi @zherman0 , The current validation behavior for value fields, where the error is shown only on clicking "Next", is not very user-friendly. Users should receive immediate feedback when they leave the value field (on blur), especially since at least one value is required for a selector. To improve usability: Show an inline validation error when the value field is left empty after user interaction (on blur). Additionally, Disable the “Add selector” action until the current selector has a valid key and at least one value. Optionally, disable “Next” until all selectors are valid.
> 
>  Screen.Recording.2026-04-30.at.7.58.28.PM.mov

I have checked above behavior. Changes LGTM @zherman0 .
@shivanthi-amara  Could you please help to update the test cases based on latest changes? 
FYI- I have assigned both day1 and day2 feature tickets to you. Thanks

shivanthi-amara (approved):
@zac Approving this PR after discussing it with the the backend team. Thank you for the efforts.

PR #477, opened 2026-04-23
OCMUI-4470 - [ROSA hosted]Machine pools status incorrectly shows "Pending" on HCP cluster overview when autoscaling min node count is zero

DcooperRH:
# Summary

This PR fixes a mismatch in the number of ready machine pools being shown now that clusters can be created with 0 min_replicas when autoscaling. 

# Jira

Fixes [OCMUI-4470](https://issues.redhat.com/browse/OCMUI-4470) 

# Additional information
Since 0 is a falsy value, doing the check 
`if (!pool.autoscaling.min_replicas || !pool.autoscaling.max_replicas)`
will always result in false if the min is 0 leading to a machine pool not being counted as ready

I replaced the check with the block below to account for this. 
`if (!hasMin(pool.autoscaling.min_replicas) || !hasMin(pool.autoscaling.max_replicas))`

We can't rely only on this check because when a cluster is initially installing, pool.status.replicas is always 0, so when the min_replicas is 0 this results in the machine pool being counted as ready right away when that is not actually the case.

I've added the below check to account for this. 
```
 if (pool.status?.message && pool.autoscaling?.min_replicas === 0) {
      return false;
    }
```

In the existing code, the message property is not being used for anything. The property displays warning message related to the status of the machine pool, such as "waitingOnMachine" or "requires minimum availability of 1..". Digging into the type definition of the property it is defined as 

```
/** @description Adds additional information about the NodePool status when the node pool doesn't reach the desired replicas. */
      message?: string;
```

Assisted by: Cursor/claude-4.6-opus
# How to Test

1. Begin creating a Rosa HCP cluster
2. When you get to the machine pools step, add 2 or 3 machine pools, enable autoscaling and set the min_replicas to 0
3. Finish creating the cluster and go to its cluster details page
4. While the cluster initially loads the Status section in the right side of the details page for machine pools should show Pending 0 / 2 or Pending 0 / 3
5. When the cluster fully builds this status sections should show Ready 2 / 2 or Ready 3 / 3. 
6. If it still shows Pending, console.log the pool value `console.log('pool', pool)` in the function `numberReadyNodePools` to check if one of the pools has the pool.status.message property with more information on why it isn't created yet

# Screen Captures

| ocmui4470_before | ocmui4470 _after |

# Review process

## QE Reviewer

[OCMUI-4470]: https://redhat.atlassian.net/browse/OCMUI-4470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/ClusterStatus.tsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/ClusterStatus.tsx
@@ -23,31 +23,34 @@ type NormalizedNodePool = Omit<NodePool, 'autoscaling'> & {
 
 // Despite being designed to handle HCP node pools, numberReadyNodePools function cannot accept type NodePool since we normalize the data to match the structure of NodePoolAutoscaling type to MachinePoolAutoscaling type.
 // See normalizeNodePool function in machinePoolsHelper.ts
-export const numberReadyNodePools = (nodePools: NormalizedNodePool[]) =>
-  nodePools?.filter((pool) => {
-    const hasMin = (min: number | undefined) => min !== undefined && min !== null;
-    const current = pool.status?.current_replicas;
+export const numberReadyNodePools = (nodePools: NormalizedNodePool[]) => {
+  const hasMin = (min: number | undefined) => min !== undefined && min !== null;
+  return (
+    nodePools?.filter((pool) => {
+      const current = pool.status?.current_replicas;
 
-    if (current === undefined) {
-      return false;
-    }
-
-    if (pool.status?.message && pool.autoscaling?.min_replicas === 0) {
-      return false;
-    }
-
-    if (pool.autoscaling) {
-      if (!hasMin(pool.autoscaling.min_replicas) || !hasMin(pool.autoscaling.max_replicas)) {
+      if (current === undefined) {
         return false;
       }
-      return current >= pool.autoscaling.min_replicas && current <= pool.autoscaling.max_replicas;
-    }
 
-    if (pool.replicas === undefined) {
-      return false;
-    }
-    return pool.replicas === current;
-  }).length || 0;
+      if (pool.status?.message && pool.autoscaling?.min_replicas === 0) {
+        return false;
+      }
+
+      if (pool.autoscaling) {
+        if (!hasMin(pool.autoscaling.min_replicas) || !hasMin(pool.autoscaling.max_replicas)) {
+          return false;
+        }
+        return current >= pool.autoscaling.min_replicas && current <= pool.autoscaling.max_replicas;
+      }
+
+      if (pool.replicas === undefined) {
+        return false;
+      }
+      return pool.replicas === current;
+    }).length || 0
+  );
+};
 
 interface ClusterStatusProps {
   cluster: ClusterFromSubscription;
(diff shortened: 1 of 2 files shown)


src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/ClusterStatus.tsx
```diff
@@ -25,13 +25,19 @@ type NormalizedNodePool = Omit<NodePool, 'autoscaling'> & {
 // See normalizeNodePool function in machinePoolsHelper.ts
 export const numberReadyNodePools = (nodePools: NormalizedNodePool[]) =>
   nodePools?.filter((pool) => {
+    const hasMin = (min: number | undefined) => min !== undefined && min !== null;
```
lizagilman: This helper function should live outside of the filter, otherwise it recreates the function on every iteration.

lizagilman (changes requested):

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/ClusterStatus.tsx
DcooperRH: Oops good catch. Fixed this in my last commit

lizagilman (approved):

jmekkatt:
Hi @DcooperRH , I have inline scenario failing at the moment. 
A ROSA HCP cluster initially has two machine pools (autoscaling enabled) with min=0 and max=1, and the cluster status is Ready (2/2). A new machine pool is added with min=2 and max=2, after which the status transitions from Pending (2/3) to Ready (3/3). When the newly added machine pool is later updated to min=0 and max=0, the cluster status changes to Pending (2/3) and does not return to a ready state. It seems like an issue and status get stuck. Could you please check this behavior?
Note: if the new machine pool is deleted, the machine pool status immediately updates from Pending (2/3) to Ready (2/2).

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/ClusterStatus.tsx
```diff
@@ -23,25 +23,34 @@ type NormalizedNodePool = Omit<NodePool, 'autoscaling'> & {
...
-  }).length || 0;
+      if (pool.status?.message && pool.autoscaling?.min_replicas === 0) {
+        return false;
+      }
+
+      if (pool.autoscaling) {
+        if (!hasMin(pool.autoscaling.min_replicas) || !hasMin(pool.autoscaling.max_replicas)) {
+          return false;
+        }
+        return current >= pool.autoscaling.min_replicas && current <= pool.autoscaling.max_replicas;
```
t-hendricks, line 44: @DcooperRH I tested the same scenario that @jmekkatt mentioned and I experienced the same results. However, based on this condition, it seems to be working as intended. Or is there a different expectation that I'm not understanding?

Image

DcooperRH: Thanks @t-hendricks . I've tested the same and I believe it is intended based on what data we can gather from the api. It looks that the current replicas hasn't updated yet.

jmekkatt: @DcooperRH Hi, is it a backend issue? I'm wondering why this happen only in above scenario.

DcooperRH: @jmekkatt see this thread https://redhat-internal.slack.com/archives/C09TPKLB3PX/p1777402511003369
We're going to need to address the 0 max issue in the UI as well

DcooperRH:
@jmekkatt let's wait for https://github.com/RedHatInsights/uhc-portal/pull/462 to merge first, but I think this will be resolved due to the fact that the max replicas field will not be able to scale down to 0

jmekkatt (approved):
Changes LGTM. Thank you.

t-hendricks (approved):

PR #498, opened 2026-04-28
OCMUI-4307: Make create cluster more prominent on OCM Overview page

meirinetta:
# Summary

This PR makes creating cluster more prominent on OCM Overview page by adding 2 buttons to the page header:
1. Create Cluster - redirects to https://console.redhat.com/openshift/create
2. Create cluster with Assisted Installer - redirects to https://console.redhat.com/openshift/assisted-installer/clusters/~new

# Jira

Fixes [OCMUI-4307](https://redhat.atlassian.net/browse/OCMUI-4307)

# Additional information

I placed the buttons within the subtitle prop because there is no built in area for buttons in this header component, only within the subtitle they can be correctly aligned (children and actionMenu do not work in terms of the layout).

# How to Test

1. Go to Overview page
2. Make sure the header looks like the [mock](https://www.figma.com/design/bSZL4vpVvGuD8xQoadj6EU/Assisted-installer-path?node-id=145-806&t=vzHVlOEEO41T7ENu-1)
3. Click on Create Cluster and Create cluster with Assisted Installer to verify the correct page is rendered
4. Try different browser widths and let me know if you notice any layout shifts.

# Screen Captures

| image | image |

# Review process

## QE Reviewer

[OCMUI-4307]: https://redhat.atlassian.net/browse/OCMUI-4307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/playwright/e2e/overview/overview.spec.ts
+++ b/playwright/e2e/overview/overview.spec.ts
@@ -6,7 +6,7 @@ test.describe.serial('OCM Overview Page tests (OCP-65189)', { tag: ['@smoke', '@
     // Navigate to overview page and wait for it to load
     await navigateTo('overview');
   });
-  test('OCM Overview Page - header and central section', async ({ overviewPage }) => {
+  test('OCM Overview Page - header and central section', async ({ overviewPage, navigateTo, page }) => {
     // Verify we're on the overview page
     await overviewPage.isOverviewPage();
 
@@ -17,11 +17,12 @@ test.describe.serial('OCM Overview Page tests (OCP-65189)', { tag: ['@smoke', '@
     );
 
     // Check Create cluster header buttons redirect properly
-    await expect(overviewPage.headerCreateClusterButton()).toHaveAttribute('href', '/openshift/create');
-    await expect(overviewPage.headerCreateClusterWithAssistedInstallerButton()).toHaveAttribute(
-      'href',
-      '/openshift/assisted-installer/clusters/~new',
-    );
+    await overviewPage.headerCreateClusterButton().click();
+    await expect(page).toHaveURL(/\/openshift\/create/);
+    await navigateTo('overview');
+    await overviewPage.headerCreateClusterWithAssistedInstallerButton().click();
+    await expect(page).toHaveURL(/\/openshift\/assisted-installer\/clusters\/~new/);
+    await navigateTo('overview');
 
     // Verify central section has expected number of cards
     await overviewPage.centralSectionCardsExpected(7);
--- a/playwright/e2e/overview/overview.spec.ts
+++ b/playwright/e2e/overview/overview.spec.ts
@@ -16,6 +16,13 @@ test.describe.serial('OCM Overview Page tests (OCP-65189)', { tag: ['@smoke', '@
       docLinks.WHAT_IS_OPENSHIFT,
     );
 
+    // Check Create cluster header buttons redirect properly
+    await expect(overviewPage.headerCreateClusterButton()).toHaveAttribute('href', '/openshift/create');
+    await expect(overviewPage.headerCreateClusterWithAssistedInstallerButton()).toHaveAttribute(
+      'href',
+      '/openshift/assisted-installer/clusters/~new',
+    );
+
     // Verify central section has expected number of cards
     await overviewPage.centralSectionCardsExpected(7);
 
--- a/src/components/overview/components/OverviewProductBanner.tsx
+++ b/src/components/overview/components/OverviewProductBanner.tsx
@@ -1,7 +1,11 @@
 import React from 'react';
 
 import PageHeader from '@patternfly/react-component-groups/dist/dynamic/PageHeader';
-import { Grid } from '@patternfly/react-core';
+import { Flex, FlexItem, Grid, Stack, StackItem } from '@patternfly/react-core';
+
+import { Link } from '~/common/routing';
+import ExternalLink from '~/components/common/ExternalLink';
+import InternalTrackingLink from '~/components/common/InternalTrackingLink';
 
 import './OverviewProductBanner.scss';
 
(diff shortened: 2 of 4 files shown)


mmenestr:
This looks great!! Just have one small comment -- can you make the c in the "Create cluster" button lower case? It should follow sentence case rules based on our UX writing guidelines :)

meirinetta:
> This looks great!! Just have one small comment -- can you make the c in the "Create cluster" button lower case? It should follow sentence case rules based on our UX writing guidelines :)

NP, Fixed

src/components/overview/components/OverviewProductBanner.tsx
```diff
@@ -21,23 +25,59 @@ export const OverviewProductBanner = ({
...
+
+  const headerActions = (
+    <Flex>
+      <FlexItem>
+        <InternalTrackingLink
+          isButton
+          variant="primary"
+          to={createClusterURL}
+          data-testid="create-cluster"
+          component={Link}
```
LiorKGOW, line 40: Although there is no direct functionality change, for consistency reasons, I would consider creating a dedicated link component like [CreateClusterLink](https://github.com/meirinetta/uhc-portal/blob/main/src/components/overview/Overview.tsx#L45-L49) as I've seen several uses similar to this one around the application.

We could also pass it down from <Overview /> and reuse logic 😄 - at least for cluster creation button. For AI we need to create a corresponding one.

Not related to this PR, Although we could work on creating a generic one, and reuse logic in all of those occurrences, I would advise doing it on a different PR/ticket, if you think that would be necessary 😇

LiorKGOW (commented):
Tested locally and everything looks great! Had one suggestion regarding the code. Let me know what you think

src/components/overview/components/OverviewProductBanner.tsx
meirinetta: @LiorKGOW Actually [CreateClusterLink](https://github.com/meirinetta/uhc-portal/blob/main/src/components/overview/Overview.tsx#L45-L49) component is redundant here and can be removed.
InternalTrackingLink already handles thelLink logic internally. 
By passing a custom component, we are redefining what the component already does.
Even though it's out of scope for this PR it's a quick change to make the code consistent (removing the CreateClusterLink component), let me know what you think.

dtaylor113:
Will review this afternoon

dtaylor113 (approved):
Tested locally.  Was wondering if the AI button should instead goto https://prod.foo.redhat.com:1337/openshift/create/datacenter where "Create cluster" via Assisted Installer is shown at the top of the page with a brief intro, but probably want a direct link to the wizard from the Overview page.
Code LGTM as well.

src/components/overview/components/OverviewProductBanner.tsx
LiorKGOW: Sounds good to me
If you think they are redundant, I would consider opening a ticket to refactor all those Links, similar to CreateClusterLink, out of the system. 

It is out of scope for this PR

LiorKGOW (approved):
Code looks good to me, and functionality works 👍🏼
Left one inline comment responding on a previous conversation we had
Thanks for the PR @meirinetta !

meirinetta:
> Tested locally. Was wondering if the AI button should instead goto https://prod.foo.redhat.com:1337/openshift/create/datacenter where "Create cluster" via Assisted Installer is shown at the top of the page with a brief intro, but probably want a direct link to the wizard from the Overview page. Code LGTM as well.

Yes, also verified this with @mmenestr, it should be directed to the wizard.

jmekkatt (changes requested):
Hi @meirinetta , 
could you please cover the newly added definitions in the existing Playwright E2E test spec : [overview.spec.ts](https://github.com/RedHatInsights/uhc-portal/blob/main/playwright/e2e/overview/overview.spec.ts)?
The key checkpoint is to verify that clicking the new buttons navigates to the correct page.
Thank you.

meirinetta:
> Hi @meirinetta , could you please cover the newly added definitions in the existing Playwright E2E test spec : [overview.spec.ts](https://github.com/RedHatInsights/uhc-portal/blob/main/playwright/e2e/overview/overview.spec.ts)? The key checkpoint is to verify that clicking the new buttons navigates to the correct page. Thank you.

Hey @jmekkatt, I added the buttons to the e2e test spec, please check the tests pass, thanks.

playwright/e2e/overview/overview.spec.ts
```diff
@@ -16,6 +16,13 @@ test.describe.serial('OCM Overview Page tests (OCP-65189)', { tag: ['@smoke', '@
       docLinks.WHAT_IS_OPENSHIFT,
     );
 
+    // Check Create cluster header buttons redirect properly
+    await expect(overviewPage.headerCreateClusterButton()).toHaveAttribute('href', '/openshift/create');
+    await expect(overviewPage.headerCreateClusterWithAssistedInstallerButton()).toHaveAttribute(
+      'href',
+      '/openshift/assisted-installer/clusters/~new',
+    );
+
```
jmekkatt, line 26: Hi @meirinetta , I think it would be better to verify the actual redirection behavior rather than checking the href properties. Please refer to the inline suggestion.

```javascript
    await overviewPage.headerCreateClusterButton().click();
    await expect(page).toHaveURL(/\/openshift\/create/);
    await navigateTo('overview');
    await overviewPage.headerCreateClusterWithAssistedInstallerButton().click();
    await expect(page).toHaveURL(/\/openshift\/assisted-installer\/clusters\/~new/);
    await navigateTo('overview');
```
WDYT?

meirinetta: @jmekkatt Yes, it does make sense, changed in the last commit

jmekkatt (approved):
Changes (including e2e tests) LGTM.

PR #517, opened 2026-05-12
OCMUI-4376: refactor MachineTypeSelection to pull state up

eliranmal:
# Summary

rewrite MachineTypeSelection as a presentation-only component.

planned and implemented with the help cursor, and updated iteratively to refactor and add fixes.

# Jira

addresses [OCMUI-4376][1]

# Additional information

this is a followup to [OCMUI-4031](https://redhat.atlassian.net/browse/OCMUI-4031), continuing the refactor made to unify the machine-type-selection components and align with typescript types.  
it is opened to address auxiliary feedback on #360 (see [this thread](https://github.com/RedHatInsights/uhc-portal/pull/360#discussion_r2841067474)).

the general aim is to pull any internal state out of the component, and make it as ‘dumb’ as possible, only rendering data.

specifically, the state to pull out is any global-state retrievals of data, as well as any leftover domain-specific logic (e.g. specific only to ROSA, or for cluster-details).

### changes

- fetch default flavor and region machine-types with react-query instead of redux, and load data via domain-specific hooks
- SoC: use a product-level hook for data, quota, effects, Formik wiring, and keep the MachineTypeSelection component strictly presentational
- add test coverage, add fixtures for missing machine-types to align with flavors

# How to Test

refer to the similar section in PR #360 (OCMUI-4031)

# Review process

## QE Reviewer

[1]: https://issues.redhat.com/browse/OCMUI-4376

[OCMUI-4376]: https://redhat.atlassian.net/browse/OCMUI-4376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OCMUI-4031]: https://redhat.atlassian.net/browse/OCMUI-4031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lizagilman:
@eliranmal what is the state of this PR? Is anything missing here? Anything not working as expected? Please elaborate so reviewers will have more context. Thanks!

eliranmal:
> @eliranmal what is the state of this PR? Is anything missing here? Anything not working as expected? Please elaborate so reviewers will have more context. Thanks!

@lizagilman - i've already put a banner at the desc' top with a note on the state of the PR and a disclaimer on what's missing.

i've now also added some text under 'additional information', for extra context (and moved the disclaimer there) ✏️

kelvah (commented):
@eliranmal I'm confused about the approach you took here. Could you explain it?

It feels like a lot of logic was moved inside `useMachineTypeSelection`, which outputs the props to feed to the shared `MachineTypeSelection`. Who owns this hook? It's part of the shared component but presentational components should know as little as possible about domain logic. In this case the hook is determining which API response to use while getting access to a lot of different information.

Part of the conversion on the other PR was also about trying to consolidate the calls into possibly a single react query instead of keeping multiple queries/sources.

From an implementation perspective, the amount of args of the custom hook is also staggering, it hints to a leaky abstraction.

eliranmal:
@kelvah, thanx for the feedback!

> Part of the conversion on the other PR was also about trying to consolidate the calls into possibly a single react query instead of keeping multiple queries/sources.

that wasn't my understanding, but using a single query makes some sense.  i'll try to create a single query that outputs all the data for machine-types and see if i can make it work (there are still data from redux store that's used in that hook, which might make this more difficult).

eliranmal:
> Who owns this hook? It's part of the shared component but presentational components should know as little as possible about domain logic

@kelvah, not sure i understand - WDYM by "who owns"?  

it's not part of the machine types component, it just handles state for machine types data.  the presentational component still only renders the data.

kelvah:
> > Who owns this hook? It's part of the shared component but presentational components should know as little as possible about domain logic
> 
> @kelvah, not sure i understand - WDYM by "who owns"?
> 
> it's not part of the machine types component, it just handles state for machine types data. the presentational component still only renders the data.

It has the same name of the machine types component with the `use` prefix. It's strictly coupled to it as it returns its props, there are no other usages of the hook that are not paired to the component. I would assume to see it in the same folder, but then this exposes the amount of complexity that still is bound to a presentational component. It's a flag responsibilities are not properly split.

Side note, also both the component and the hook live inside `ScaleSection` while the component is used in different places, it should be moved to the top level `components/common` folder.

eliranmal:
> Part of the conversion on the other PR was also about trying to consolidate the calls into possibly a single react query instead of keeping multiple queries/sources.

@kelvah , IRT consolidating calls into a single query, this is becoming a huge change.   in the current impl' (before this PR) machine-types data is already passed around many layers in the component tree, so attempting to consolidate it would result in changing probably around 30 additional files..  (if it's even possible).

this pattern of passing down props with data is deeply rooted in the codebase (see EditDetailsSection, EditMachinePoolModal, and upwards to MachinePools), making the code so rigid, that this change is probably out of scope for this PR.

OTOH, i'm seeing an existing pattern of hooks dedicated for component-specific needs..  not sure it's confirmation that this is a good thing, of course  :)  
see for example `useMachinePoolFormik`, `useLabelsTagsTaintsSubTab`, `useSecurityGroupsSubTab`, etc.

PR #533, opened 2026-05-19
OCMUI-4044: cleanup insights-proxy

eliranmal:
# Summary

remove any insights-proxy related code and libraries as obsolete.  we now use `fec` for a dev-env, which spins up a proxy internally.

# Jira

resolves [OCMUI-4044]

# Additional information

### changes

- remove these scripts as obsolete:
  - insights-proxy
  - wait-on-insights-proxy
  - check-spandex-port-available
- remove the obsolete package `porty` (was only used inside check-spandex-port-available)
- update makefile command for dev-env-setup to run `fec patch-etc-hosts` instead of running the script from insights-proxy repo (and rename tasks for semantics)
- remove obsolete npm scripts (`insights-proxy`, `stop-insights-proxy`), and update the `stop` script to remove the insights-proxy stop command
- update readme to add an arguments separator to the fec patch command
- off topic:  update makefile's `setup` task to use `npm install` instead of `npm ci` (`ci` [always removes](https://docs.npmjs.com/cli/v8/commands/npm-ci#description) `node_modules`, but makefile commands are mainly used for local dev' and not in CI, so a clean install isn't suitable for that)

# How to Test

- run `make dev-env-setup` locally and ensure it has the same result (adds localhost mappings to the /etc/hosts file, when they're missing)
- run `npm start` locally and make sure the dev-server works the same as before

# Review process

## QE Reviewer

no feature testing.

[OCMUI-4044]: https://redhat.atlassian.net/browse/OCMUI-4044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Makefile
```diff
@@ -66,20 +60,13 @@ insights-proxy-check: run/insights-proxy
...
 
 .PHONY: dev-env-setup
-dev-env-setup: run/insights-proxy
-	sudo bash -x run/insights-proxy/scripts/patch-etc-hosts.sh
-
-.PHONY: insights-proxy-setup
-insights-proxy-setup: dev-env-setup
-	run/podman-or-docker.sh pull quay.io/redhat-sd-devel/insights-proxy:3.2.1
+dev-env-setup: node_modules
+	sudo npm exec -- fec patch-etc-hosts
```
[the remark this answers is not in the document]
eliranmal: hmm, but:
- the prev' command also used `sudo`
- this makefile command is used in the scope of CI - how can we leave `fec patch-etc-hosts` to request elevation directly when required, when the command will always be hard-coded from CI POV?

DcooperRH (approved):

README.md
```diff
@@ -25,7 +25,7 @@ _\* see package.json for the compatible versions_
 
 ## Setup
 
-For a first time setup, it's required to run `npm exec fec patch-etc-hosts`.
+For a first time setup, it's required to run `npm exec -- fec patch-etc-hosts`.
```
lizagilman, line 28:
```suggestion
For a first time setup, it's required to run `sudo npm exec -- fec patch-etc-hosts`.
```
It does actually need `sudo`, no? 

https://github.com/RedHatInsights/uhc-portal/pull/533/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R64

lizagilman (approved):

README.md
eliranmal: it might\*, and the following paragraph in that doc states that.

(\* it's only certain that it's needed in **CI** – where the user is guaranteed to not be a root user or in the sudoers group, and the shell is non-interactive)

Makefile
```diff
@@ -16,22 +16,16 @@
 
 # Default target
 .PHONY: setup
-setup: node_modules insights-proxy-check
+setup: node_modules dev-env-check
 
 .PHONY: node_modules
 node_modules:
-	npm ci
+	npm install
```
jmekkatt, line 23: It looks like `npm ci` installs dependencies exactly from package-lock.json and fails if it doesn't match package.json.
But `npm install` can automatically update package-lock.json when differences are found. Because of this, anyone running `make setup` or `make node_modules` could accidentally change and commit the lockfile, affecting builds.
WDYT @eliranmal ?

eliranmal: there's no risk of changing the lockfile unless you changed your package.json locally – and in which case, you'd want the lockfile to change anyway.

(this script isn't used in CI, but only meant for local setup.  that change to `npm ci` was wrongfully done within the yarn/npm migration PR – i mentioned this in the PR desc').

shivanthi-amara (approved):
LGTM

PR #536, opened 2026-05-19
OCMUI-4525 - [Cluster Details -> Overview tab] Display AutoNode status

DcooperRH:
# Summary

This PR adds the ability to enable Autonode for ROSA HCP clusters. A new Autonode field is added to the Cluster details overview where a user can open a new modal to configure the Autonode settings. A new Karpenter nodes count field has been added as well underneath the original Nodes (compute) count 

Assisted by: Cursor/claude-opus-4-6

# Jira

Fixes [OCMUI-4525](https://issues.redhat.com/browse/OCMUI-4525) 

# Additional information

I also updated the reusable component ErrorBox.tsx to add padding bottom to the component. The style from app.scss was not getting picked up. I've noticed across the app that the padding bottom is missing when this component is rendered. 

A follow-up task has been created here to add a "Learn more" link for the field https://redhat.atlassian.net/browse/OCMUI-4567

# How to Test

1. Create a ROSA HCP cluster, but use version nightly or candidate of 4.22 or higher. This feature requires version 4.22 or higher and we do not yet have it available as a stable or eus version
2. Create an ARN that can be configured for Autonode 
```
aws iam create-role \
  --role-name ManagedOpenShift-AutoNode-Role \
  --assume-role-policy-document '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"Service":"ec2.amazonaws.com"},"Action":"sts:AssumeRole"}]}'
```
3. get the `OIDC Endpoint URL: ` by running `rosa describe cluster -c <cluster-name>`
4. Go to the AWS console --> IAM --> Roles. FInd your Autonode arn you created
5. Update the trust policy to (do not include  https:// from the OIDC Endpoint URL, for example use `d3gt1gce2zmg3d.cloudfront.net/22qa79chsq8mand8hvmnr33upj48lmas` instead of `https://d3gt1gce2zmg3d.cloudfront.net/22qa79chsq8mand8hvmnr33upj48lmas` )
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "Federated": "arn:aws:iam::720424066366:oidc-provider/<oidc-endpoint-url>"
            },
            "Action": "sts:AssumeRoleWithWebIdentity",
            "Condition": {
                "StringEquals": {
                    "<oidc-endpoint-url>:sub": "system:serviceaccount:kube-system:karpenter"
                }
            }
        }
    ]
}
```
6. In your role's permissions add the json policy found here https://gitlab.cee.redhat.com/service/ocm-backend-tests/-/blob/master/templates/autonode/autonode-iam-policy.json?ref_type=heads (vpn required)
7. Open the Autonode modal through the pencil icon. Enable Autonode, note that once Autonode is enabled and the modal is submitted, it cannot be disabled. 
8. After enabling and saving the modal, the pencil icon tooltip should not be disabled as you should be allowed to update the Arn field again, but you cannot disable Autonode.

NOTE: The value from DetailsRight.tsx --> `const autoNodeCount = cluster?.auto_node?.status?.node_count;` is not yet accessible as the change to add this key has not yet been deployed by the backend team. (ETA May21, or May22). For now the _Autonode (Karpenter):_ node count field will not show because we don't have access to the value. If you want to test the display of this field, you can change the variable autoNodeCount to `const autoNodeCount = 5` (or any number)
# Screen Captures

| Autonode modal                                              | 
| --------------------------------------------------- | 
| ocmui4525_modal | 

| Karpenter nodes                                              | 
| --------------------------------------------------- | 
| ocmui4525_nodesDisplay | 

| Autonode display when enabled                                            | When Autonode has error status message                                  |
| ocmui4525_display | ocmui4525_autonodeErrorstatus |

# Review process

## QE Reviewer

[OCMUI-4525]: https://redhat.atlassian.net/browse/OCMUI-4525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+const EditAutoNodeModal = ({ cluster, region, onClose }: EditAutoNodeModalProps) => {
+  const [isAutoNodeEnabled, setIsAutoNodeEnabled] = React.useState(
+    cluster?.auto_node?.mode === 'enabled',
+  );
+  const [iamRoleArn, setIamRoleArn] = React.useState(cluster?.aws?.auto_node?.role_arn ?? '');
+  const [arnValidationError, setArnValidationError] = React.useState<string | undefined>(undefined);
+
+  const isValid = !isAutoNodeEnabled || (iamRoleArn.trim().length > 0 && !arnValidationError);
+
```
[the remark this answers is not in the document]
DcooperRH: You can't pre-load an invalid arn as you must get passed this validation to initially submit the modal

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
```diff
@@ -405,6 +453,53 @@ function DetailsRight({ cluster, hasAutoscaleCluster, isDeprovisioned, clusterDe
...
+                buttonAriaLabel="More information about Autonode"
+                hint="Enables hosted Karpenter to autoscale nodes."
+              />
+            </DescriptionListTerm>
+            <DescriptionListDescription>
+              <EditButton
+                data-testid="editAutoNodeButton"
+                ariaLabel="Edit Autonode settings"
+                disableReason={
+                  (!cluster?.canEdit && 'You do not have permission to edit Autonode settings.') ||
```
lizagilman, line 495:
```suggestion
                  (!cluster?.canUpdateClusterResource && 'You do not have permission to edit Autonode settings.') ||
```

lizagilman (changes requested):

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
```diff
@@ -287,6 +319,14 @@ function DetailsRight({ cluster, hasAutoscaleCluster, isDeprovisioned, clusterDe
...
                   </dd>
                 </Flex>
+                {isAutoNodeAllowed && autoNodeCount != null && (
+                  <Flex data-testid="autoNodeKarpenterCountContainer">
+                    <dt>Autonode (Karpenter): </dt>
+                    <dd data-testid="autoNodeKarpenterCount">
+                      {autoNodeCount} {autoNodeKarpenterHint}
+                    </dd>
+                  </Flex>
+                )}
```
lizagilman, line 339: DRY: consider extracting it to a component so it can be reused in Lines 354-361

jmekkatt (changes requested):
Hi @DcooperRH AutoNode settings could be editable when cluster is not ready state ex: installing. The AutoNode option should be disabled and shouldn't allow user to change the configuration when cluster is not ready.
Refer attached recording

https://github.com/user-attachments/assets/d1ddd062-156d-41f0-8225-b40538a9d919

jmekkatt (commented):
Save button always enabled for an already-disabled cluster with no changes

Steps to reproduce

1. Open the Autonode modal on a cluster with Autonode disabled.
2. Do not change anything.
3. Click Save.
4. A PATCH request is sent to the cluster API with the same disabled state.

Expected behaviour 
If no fields have changed, the Save button should be disabled, or the handler should compare initial vs. current state and skip the API call if they are identical.

Actual behaviour 
Save is always enabled when Autonode is disabled; every modal open-and-close cycle fires a redundant PATCH.

https://github.com/user-attachments/assets/c252b15a-bf49-4950-8020-11211c4f2558

jmekkatt (commented):
Asterisk shown unconditionally even when ARN field is disabled
The AutoNode IAM role ARN field always renders as required-field with asterisk regardless of whether Autonode is enabled or disabled. When Autonode is disabled the ARN field is disabled and is not required for form submission. Showing isRequired on a disabled, non-required field is contradictory and misleads users.

Steps to reproduce

1. Open the Autonode modal on a cluster with Autonode disabled.
2. Observe the ARN field is greyed out (disabled) but still shows a red asterisk (*) marking it as required.

jmekkatt:
Hi @DcooperRH , I'm seeing an issue with autonode status on an enabled auto node configuration. The Autonode status shown as  "AutoNode provisioner is not configured" . I couldn't get much information about it from backend calls. 
Could you please confirm what could be missing from my side?
Screenshot 2026-05-26 at 5 51 21 PM

jmekkatt (commented):
Hi @DcooperRH , Currently, the "Edit autonode settings" dialog does not indicate why the "Enable autonode" toggle is disabled once autonode configuration has already been enabled. It would be helpful to add a tooltip explaining the reason, so users can better understand why the toggle cannot be modified.

Image

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
```diff
@@ -141,6 +158,21 @@ function DetailsRight({ cluster, hasAutoscaleCluster, isDeprovisioned, clusterDe
   const showDeleteProtection = cluster.managed && !isArchivedSubscription(cluster);
   const isClusterUninstalling = cluster.state === clusterStates.uninstalling;
 
+  const autoNodeKarpenterHint = (
+    <PopoverHint
+      iconClassName="nodes-hint"
+      buttonAriaLabel="More information about AutoNode Karpenter nodes"
+      hint="These nodes are automatically provisioned and managed by Karpenter based on workload demands. These nodes are not managed through machine pools."
+    />
+  );
```
t-hendricks: DRY: this hint can be extracted to the same component that @lizagilman [suggested](https://github.com/RedHatInsights/uhc-portal/pull/536#discussion_r3296887893)

DcooperRH:
@jmekkatt @lizagilman @t-hendricks Thank you for the first pass through on the review. I believe all issues raised have been addressed

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/AutoNodeKarpenterCount.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <dt>Autonode (Karpenter): </dt>
+    <dd data-testid="autoNodeKarpenterCount">
+      {count}{' '}
+      <PopoverHint
+        iconClassName="nodes-hint"
+        buttonAriaLabel="More information about AutoNode Karpenter nodes"
+        hint="These nodes are automatically provisioned and managed by Karpenter based on workload demands. These nodes are not managed through machine pools."
+      />
+    </dd>
+  </Flex>
```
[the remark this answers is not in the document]
DcooperRH: This is following the same format as the nodes fields displayed above this component so that they are uniform

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/AutoNodeKarpenterCount.tsx
```diff
@@ -0,0 +1,25 @@
...
+};
+
+const AutoNodeKarpenterCount = ({ count }: AutoNodeKarpenterCountProps) => (
+  <Flex data-testid="autoNodeKarpenterCountContainer">
+    <dt>Autonode (Karpenter): </dt>
+    <dd data-testid="autoNodeKarpenterCount">
+      {count}{' '}
+      <PopoverHint
+        iconClassName="nodes-hint"
+        buttonAriaLabel="More information about AutoNode Karpenter nodes"
```
lizagilman: Naming inconsistency — the Jira comments from Bala confirmed we should avoid camelCase `AutoNode` and use `Autonode` instead. This aria label should be `"More information about Autonode Karpenter nodes"`. The corresponding assertion in `AutoNodeKarpenterCount.test.tsx` will need to be updated to match.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.tsx
```diff
@@ -0,0 +1,170 @@
...
+              onChange={(_event, checked) => setIsAutoNodeEnabled(checked)}
+              isDisabled={cluster?.auto_node?.mode === 'enabled'}
+            />
+          </WithTooltip>
+
+          {isAutoNodeEnabled && cluster?.auto_node?.mode !== 'enabled' ? (
+            <Alert variant="warning" isInline isPlain title="Enabling Autonode cannot be undone">
+              Ensure you have created the required IAM role in your AWS account before proceeding.
+            </Alert>
+          ) : null}
```
lizagilman, line 130: Same naming issue — `"AutoNode IAM role ARN"` should be `"Autonode IAM role ARN"` (lowercase 'n', per Jira branding guidance).

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
```diff
@@ -62,6 +70,15 @@ function DetailsRight({ cluster, hasAutoscaleCluster, isDeprovisioned, clusterDe
 
   const [hasAutoscaleMachinePools, setHasAutoscaleMachinePools] = React.useState();
   const [limitedSupport, setLimitedSupport] = React.useState();
+  const [isEditAutoNodeModalOpen, setIsEditAutoNodeModalOpen] = React.useState(false);
+
+  const AUTO_NODE_MIN_VERSION = '4.22.0';
+  const clusterVersion = cluster?.openshift_version || cluster?.version?.raw_id || '';
+
+  const coercedVersion = semver.coerce(clusterVersion);
+  const isAutoNodeVersionValid = semver.coerce(coercedVersion)
```
lizagilman: `semver.coerce(coercedVersion)` is called on a value that is already the result of `semver.coerce()` — `coercedVersion` is already a `SemVer | null`. The check should just be:
```js
const isAutoNodeVersionValid = coercedVersion
  ? semver.gte(coercedVersion, AUTO_NODE_MIN_VERSION)
  : false;
```

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
```diff
@@ -62,6 +70,15 @@ function DetailsRight({ cluster, hasAutoscaleCluster, isDeprovisioned, clusterDe
 
   const [hasAutoscaleMachinePools, setHasAutoscaleMachinePools] = React.useState();
   const [limitedSupport, setLimitedSupport] = React.useState();
+  const [isEditAutoNodeModalOpen, setIsEditAutoNodeModalOpen] = React.useState(false);
+
+  const AUTO_NODE_MIN_VERSION = '4.22.0';
```
lizagilman: `AUTO_NODE_MIN_VERSION` is a true constant — it should be defined at module scope rather than inside the component function so it isn't recreated on every render.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  const [arnValidationError, setArnValidationError] = React.useState<string | undefined>(undefined);
+
+  const isValid = !isAutoNodeEnabled || (iamRoleArn.trim().length > 0 && !arnValidationError);
+  const hasChanges =
+    isAutoNodeEnabled !== initialAutoNodeEnabled || iamRoleArn !== initialIamRoleArn;
+
+  const { mutate: editCluster, isPending: isSubmitting, isError, error } = useEditCluster(region);
+
+  const handleSave = () => {
```
lizagilman, line 57: The Jira AC lists two Segment events that are missing:
- `Autonode Enable Modal Opened` — should fire when the modal is opened (in the `onClick` in `DetailsRight`)
- `Autonode Enable Submitted` — should fire here in `handleSave` on a successful mutation

Other modals in the codebase fire Segment events for equivalent actions.

src/components/common/ErrorBox.tsx
```diff
@@ -62,7 +62,7 @@ const ErrorBox = ({
       isInline
       title={message}
       role={variant === 'danger' || forceAsAlert ? 'alert' : undefined}
-      className="error-box"
+      className="pf-v6-u-mb-sm"
```
lizagilman: This change replaces `className="error-box"` (which was scoped to modals via `.pf-v6-c-modal-box .error-box` in `app.scss`) with `className="pf-v6-u-mb-sm"`, which applies a margin-bottom to **every** `ErrorBox` render across the entire app — not just inside modals.

The old `app.scss` rule has also been removed. If the intent is to fix modal spacing, a narrower approach (e.g. a rule on `.pf-v6-c-modal-box__footer`) would avoid unintended side effects on non-modal uses of `ErrorBox`.

lizagilman:
@DcooperRH posted an AI-assited code review comments. I reviewed them before posting (actually I dismissed some issues AI flagged), the remaining comments looks reasonable. PTAL, thanks.

t-hendricks (commented):
Found a bug in the modal where the `Save` button is enabled while the `Enable Autonode` button is disabled after adding text in the `TextInput` field.

https://github.com/user-attachments/assets/7c631629-361e-42a7-ba8e-664bf5b0e0f9

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/AutoNodeKarpenterCount.tsx
DcooperRH: Good catch, I fixed this

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.tsx
DcooperRH: idk how I missed the segment tracking, but I've added it now

src/components/common/ErrorBox.tsx
DcooperRH: Ok. I added the below to app.scss that seems to be working correctly 
```
.pf-v6-c-modal-box__footer .error-box {
  margin-bottom: 15px;
}
```

DcooperRH:
> Found a bug in the modal where the `Save` button is enabled while the `Enable Autonode` button is disabled after adding text in the `TextInput` field.
> 

@t-hendricks Thank you for finding this. I just pushed a change to fix the issue

src/styles/app.scss
```diff
@@ -590,12 +594,17 @@ $ocm-c-width-limiter--breakpoint-map: build-breakpoint-map();
 }
 
 // Fix selectable card border issue due to a style override in chroming
-:is(.ocm-fix-selectable-card-border .pf-v6-c-card, .ocm-fix-selectable-card-border.pf-v6-c-card):is(.pf-m-clickable, .pf-m-selectable)::before {
```
DcooperRH, line 593: I believe the two changes below here are from running prettier on app.scss

jmekkatt:
> Found a bug in the modal where the `Save` button is enabled while the `Enable Autonode` button is disabled after adding text in the `TextInput` field.
> 
>  Screen.Recording.2026-06-01.at.5.58.52.PM.mov

In addition to @t-hendricks' findings, I think we should clear the entered ARN value and any ARN validation errors when the user disables the Autonode feature. 
@DcooperRH Hi, Kindly check the below recording 
https://github.com/user-attachments/assets/6a8fd73a-a090-4fd4-ac9f-3bf352dbec99

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/AutoNodeKarpenterCount.tsx
jmekkatt: The ROSA CLI seems to be camelCase and not aligned with UI with above changes.
Screenshot 2026-06-02 at 2 25 16 PM

Probably we should know if the naming convention is only specific to OCMUI. 
@DcooperRH Could you please check?

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
```diff
@@ -141,6 +163,13 @@ function DetailsRight({ cluster, hasAutoscaleCluster, isDeprovisioned, clusterDe
   const showDeleteProtection = cluster.managed && !isArchivedSubscription(cluster);
   const isClusterUninstalling = cluster.state === clusterStates.uninstalling;
 
+  const autoNodeWarningAlert =
+    isAutoNodeAllowed && cluster?.auto_node?.status?.message ? (
+      <Alert variant="warning" isInline title="Autonode status" className="pf-v6-u-mt-sm" isPlain>
```
lizagilman, line 186: Should it be `variant="danger"` ? IMHO warning is fine, but JIRA says it should be red.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.tsx
```diff
@@ -0,0 +1,176 @@
...
+      },
+      {
+        onSuccess: () => {
+          track(trackEvents.AutonodeEnableSubmitted);
+          queryClient.invalidateQueries({
+            queryKey: [
+              queryConstants.FETCH_CLUSTER_DETAILS_QUERY_KEY,
+              'clusterService',
+              cluster.id,
+              cluster.subscription,
```
lizagilman: @DcooperRH <strike> same as with the machine pool modal issue, this will cause re-render cycle that un-mounts the modal from the view. </strike> This is not true.
<strike>@kelvah can we safely remove it from here? We discussed about trying to find a reason behind including subscription in the query key in `src/queries/ClusterDetailsQueries/useFetchCluster.ts` , but does it also apply to all other queries targeting invalidating cluster details related queries? </strike>
And this is not related.
Confused it with query keys of refetch cluster details data but it's not related. Sorry for the noise.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.tsx
```diff
@@ -0,0 +1,176 @@
...
+            ],
+          });
+          onClose();
+        },
+      },
+    );
+  };
+
+  return (
+    <Modal
```
lizagilman, line 93: Unfortunately (and similar to the recent machine pool modal issue) this modal makes our tech debt grow since it's also not guarded from auto-refresh that may unmount it from the view. We can do one of the following -
1. Ignore it for now, have a proper fix for everything together. Trade-offs: Bad UX, and as mentioned tech debt becomes even bigger.
2. Apply the same fix we applied to machine pool modal. Trade-off: We want to move away from redux-modals and redux entirely, this would be a step to the opposite direction.
3. Come up with a new proper pattern and gradually move all modals there. Trade-off: Till we migrate all modals we will have to maintain 2 patterns - legacy redux and the new one. Also it widens the scope of this PR/feature.
cc @kelvah

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
```diff
@@ -405,6 +440,59 @@ function DetailsRight({ cluster, hasAutoscaleCluster, isDeprovisioned, clusterDe
...
+              <EditButton
+                data-testid="editAutoNodeButton"
+                ariaLabel="Edit Autonode settings"
+                disableReason={
+                  (!cluster?.canUpdateClusterResource &&
+                    'You do not have permission to edit Autonode settings.') ||
+                  (!isAutoNodeVersionValid &&
+                    `Autonode requires OpenShift version ${AUTO_NODE_MIN_VERSION} or above.`) ||
+                  (cluster?.state !== clusterStates.ready &&
+                    'Autonode settings can only be edited when the cluster is ready.')
```
lizagilman: Are you sure it resolves to the expected type of `disableReason` in `EditButton` component?

jmekkatt:
Hi @DcooperRH , I have tried multiple ways to provision the Karpenter nodes against the ROSA HCP cluster. It looks like the node count  return as```Autonode (Karpenter): 0  ``` 
Screenshot 2026-06-02 at 6 22 09 PM

Could you please check and update the steps to provision autonodes in ROSA HCP cluster.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
DcooperRH: I think it should not be danger, as some of the status messages that are returned are not errors.

Edit: Bala has approved of the current display for this alert

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.tsx
DcooperRH: I think we should just remove it considering this is just a two field simple modal

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
DcooperRH: I believe so. There are unit tests in place to cover each disabled scenario. While testing I've seen each reason displayed correctly. Imo this is the correct order of precedence for the disabledReason

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.tsx
DcooperRH: I don't think we have a solution decided by the team ready to go regarding the modals. As proposed fix 2 is temporary, I feel like implementing it here will add to tech debt. We will likely implement a fix for all modals at the same time imo.
This modal is a pretty simple two component form, so we don't run as much risk of users being frustrated if the details refresh and the modal closes on its own compared to the machine pool modal

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/AutoNodeKarpenterCount.tsx
DcooperRH: Looks like it should be Autonode, and the main label in the cluster details is chaning to `Red Hat build of Karpenter (Autonode)` according to this [thread](https://redhat-internal.slack.com/archives/C088H9VV9UM/p1778789820297649)

DcooperRH:
> In addition to @t-hendricks' findings, I think we should clear the entered ARN value and any ARN validation errors when the user disables the Autonode feature. @DcooperRH Hi, Kindly check the below recording https://github.com/user-attachments/assets/6a8fd73a-a090-4fd4-ac9f-3bf352dbec99

@jmekkatt Fixed this

jmekkatt:
Hi @DcooperRH , As discussed in the [DDR mockups](https://docs.google.com/document/d/1_ub0mR4OM6FM45zbWXKLefaP96xSuDDQ8GdZ9BI5yfI/edit?disco=AAAB6g_jotI), Autonode configuration is a cluster-level property and is not specific to the machine pool area. Based on that, I would like to confirm whether users with the Machine Pool Editor role should be allowed to add or update this configuration.
The current implementation allows Machine Pool Editor users to add and update the Autonode configuration. Could you please review whether this behavior is intended? Thank you.

jmekkatt:
Hi @DcooperRH , Autonode IAM role ARN field has no visual error state when user specify invalid ARN

1. Open the Edit Autonode modal on a disabled cluster.
2. Enable the Autonode toggle.
3. Type "invalid-not-an-arn" into the IAM role ARN field.

Actual :  The helper text below shows an error, but the input field border remains unchanged (no red).
Screenshot 2026-06-03 at 3 10 14 PM

Expected : The Text input border turns red to visually signal an invalid value, consistent with PatternFly form validation patterns.
In other models like "Add machine pool" with invalid name in name field shows as below
Screenshot 2026-06-03 at 3 11 52 PM
Could you please update?

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.tsx
lizagilman: Ok we've been discussing it and `cluster.subscription` is not needed here, let's remove it. Our query keys are pretty confusing, we should be more careful with them.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.tsx
lizagilman: Also been discussing that, @/kelvah suggest for new modals we should apply the fix, at least until we remove the subscription from the useFecthCluster query key. Then long term we need to migrate all modals at once.

DcooperRH: Ok. I will add this once we can resume development

DcooperRH: ok I added a fix similar to the machine pool modal fix. I tested to see if the modal would close on its own and could not get it to after running a timer for 5 minutes

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.tsx
DcooperRH: Just removed it in my last push

src/components/common/ErrorBox.tsx
lizagilman: Looks like it fixes https://redhat.atlassian.net/browse/OCMUI-4314 :tada:

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.test.tsx
```diff
@@ -0,0 +1,421 @@
...
+
+      const input = screen.getByDisplayValue(
+        'arn:aws:iam::123456789012:role/ManagedOpenShift-Autonode-Role',
+      );
+      await user.clear(input);
+      await user.type(input, 'arn:aws:iam::999999999999:role/NewRole');
+
+      // Note: the switch is disabled for enabledCluster, so this test only applies
+      // if you allow toggling off in some scenario. If not, skip this test.
+    });
```
lizagilman: At the moment we don't have this scenario, there's only setup here it doesn't really test anything. Better to remove it I think.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
```diff
@@ -62,6 +79,26 @@ function DetailsRight({ cluster, hasAutoscaleCluster, isDeprovisioned, clusterDe
 
   const [hasAutoscaleMachinePools, setHasAutoscaleMachinePools] = React.useState();
   const [limitedSupport, setLimitedSupport] = React.useState();
+  const [isEditAutoNodeModalOpen, setIsEditAutoNodeModalOpen] = React.useState(false);
+
+  // Pause auto-refresh while the edit Autonode modal is open.
+  const openEditAutoNodeModal = () => {
+    track(trackEvents.AutonodeEnableModalOpened);
```
lizagilman, line 88: Do we need to track it?

lizagilman (approved):

lizagilman (approved):
Thanks for all the effort @DcooperRH !

lizagilman:
> Hi @DcooperRH , As discussed in the [DDR mockups](https://docs.google.com/document/d/1_ub0mR4OM6FM45zbWXKLefaP96xSuDDQ8GdZ9BI5yfI/edit?disco=AAAB6g_jotI), Autonode configuration is a cluster-level property and is not specific to the machine pool area. Based on that, I would like to confirm whether users with the Machine Pool Editor role should be allowed to add or update this configuration. The current implementation allows Machine Pool Editor users to add and update the Autonode configuration. Could you please review whether this behavior is intended? Thank you.

Machine pool role is too wide with permissions this is a backend issue affecting many features. It's the same issue causing https://redhat.atlassian.net/browse/ROSAENG-8210 . Actually there are some recent discussions on that, linked from the ticket.

DcooperRH:
> > Hi @DcooperRH , As discussed in the [DDR mockups](https://docs.google.com/document/d/1_ub0mR4OM6FM45zbWXKLefaP96xSuDDQ8GdZ9BI5yfI/edit?disco=AAAB6g_jotI), Autonode configuration is a cluster-level property and is not specific to the machine pool area. Based on that, I would like to confirm whether users with the Machine Pool Editor role should be allowed to add or update this configuration. The current implementation allows Machine Pool Editor users to add and update the Autonode configuration. Could you please review whether this behavior is intended? Thank you.
> 
> Machine pool role is too wide with permissions this is a backend issue affecting many features. It's the same issue causing https://redhat.atlassian.net/browse/ROSAENG-8210 . Actually there are some recent discussions on that, linked from the ticket.

@lizagilman Thank you for looking at this.. so does there need to be a fix here in this PR or should we leave the permission as is and fix the machine pool editor role issue separately? I guess it's a backend issue so I'm not sure what we can do on our side. I also don't want to add more tech debt with a workaround for this

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
t-hendricks, line 447: Missing footer under Min/Max: “_Min/Max applies to machine pool nodes only. Autonode (Karpenter) may provision additional nodes beyond this range._” when Autoscaling and Autonode is enabled.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
```diff
@@ -405,6 +456,52 @@ function DetailsRight({ cluster, hasAutoscaleCluster, isDeprovisioned, clusterDe
...
+            <EditAutoNodeModal cluster={cluster} region={region} onClose={closeEditAutoNodeModal} />
+          )}
+          <DescriptionListGroup>
+            <DescriptionListTerm>
+              Red Hat build of Karpenter (Autonode)
+              <PopoverHint
+                id="autonode-hint"
+                iconClassName="nodes-hint"
+                buttonAriaLabel="More information about Autonode"
+                hint="Enables hosted Karpenter to autoscale nodes."
```
t-hendricks, line 486: Should the `hint` be updated to "_Nodes provisioned automatically by Karpenter based on workload demand. These nodes are not managed through machine pools._" when Autonode is enabled?

t-hendricks (commented):
@DcooperRH Can you verify per Acceptance Criteria (**Nodes section (Karpenter node count)**) and the provided mockups if this is needed?

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
DcooperRH: Not for this one. The mockups may be out of date. We have a hint on the Karpenter node count that says "These nodes are automatically provisioned and managed by Karpenter based on workload demands. These nodes are not managed through machine pools."

The hint here "Enables hosted Karpenter to autoscale nodes." came directly from Bala

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
DcooperRH: When I first saw the text there in the mockup I thought for some reason it was just to help with the mockup.. good catch. I've added PFs HelperText component for this 
Screenshot 2026-06-04 at 1 33 41 PM

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.test.tsx
DcooperRH: Removed this, thank you

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
DcooperRH: It asks to track a user opening the modal in the [DDR](https://docs.google.com/document/d/1ZEWeUjXWS-8yL3PUhCNscQfeHeFEZX7GMyhNNFurW8U/edit?tab=t.0#heading=h.vgysw1m5iiia)

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
t-hendricks: Verified the change. Thanks for adding this. Resolving thread.

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
t-hendricks: Thanks for the clarification. Resolving this thread.

t-hendricks (approved):

DcooperRH:
@t-hendricks @lizagilman @jmekkatt , I've decided to update the cluster permission to cluster.canEdit since it is the desired behavior when it comes to having access to this modal. I know cluster.canUpdateResource is ideally the better approach but since it is bugged with the machine pool editor role and we do not know when it will be resolved in the backend, I feel this is the best approach in the meantime since Bala has asked that machine pool editor role cannot modify this

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
jmekkatt: @DcooperRH Could we update the acceptance criteria based on above.? Thank you.

jmekkatt (approved):
Completed pre-merge testing and changes lgtm.
 Thank you @DcooperRH

src/components/clusters/ClusterDetailsMultiRegion/components/Overview/DetailsRight/DetailsRight.jsx
DcooperRH: It is in the Jira AC under 
**Acceptance Criteria**

**Autonode status field (right panel, above Network):**

- Help icon tooltip: “Enables hosted Karpenter to autoscale nodes.”

lizagilman:
> @t-hendricks @lizagilman @jmekkatt , I've decided to update the cluster permission to cluster.canEdit since it is the desired behavior when it comes to having access to this modal. I know cluster.canUpdateResource is ideally the better approach but since it is bugged with the machine pool editor role and we do not know when it will be resolved in the backend, I feel this is the best approach in the meantime since Bala has asked that machine pool editor role cannot modify this

@DcooperRH Machine pool editor role can still modify it, you are just cosmetically disabling it, and in the same time you skip the correct permission check. It's not about better or worse approach, it will just make the UI lie by showing a disabled button for users with `canUpdateClusterResource` =  `true` and `canEdit` = `false` , which is wrong, because they have enough permissions. The other way round could also happen and will also make the UI lie - if user has `canUpdateClusterResource` =  `false` and `canEdit` = `true` - the button will be enabled while it shouldn't be , not enough permissions in this case. 

You should push back on that and tell Bala - 
1. UI must check the same permissions beckend does, otherwise it lies, resulting poor UX.
2. The wide permissions on the machine pool role are a known, long-standing backend issue, that affects many areas of the UI. It is not specific to autonode. Unfortunately, despite multiple requests from our side, the backend team has not yet prioritized a fix. It will be helpful if Bala can reach out to them directly to push for higher prioritization.

Please open a follow up PR to adjust it back.

PR #562, opened 2026-05-31
OCMUI-4623: fix auto-refresh interrupting the edit/add machine pool modal

lizagilman:
# Summary

**The issue:** While the Add/Edit machine pool modal is open, auto-refresh keeps firing and can cause the page to reload mid-edit, closing the modal and losing the user's changes.

**Root cause:**
- \`EditMachinePoolModal\` is controlled by local state (\`editMachinePoolId\` / \`addMachinePool\`), not the Redux modal system, so the existing \`anyModalOpen\` guard in \`ClusterDetails\` stays \`false\` while the modal is open
- As a result, the \`RefreshButton\` timer keeps firing and calling \`refreshClusterDetails()\`, which invalidates all queries tagged with \`FETCH_CLUSTER_DETAILS_QUERY_KEY\`, including the subscription query
- The entire \`subscription\` object is embedded in the \`useFetchCluster\` query key. When any field in it changes server-side (roughly every ~5 minutes), React Query sees it as a new query, drops the cache, and triggers a full cluster re-fetch — reloading the page and closing the modal

**Fix:** Dispatch \`openModal(modals.EDIT_MACHINE_POOL)\` / \`closeModal()\` directly in the event handlers that open and close the modal (\`openEditMachinePoolModal\`, \`openAddMachinePoolModal\`, \`closeEditMachinePoolModal\`), so the existing guard correctly pauses auto-refresh for the duration of the modal session

# Jira

Fixes [OCMUI-4623](https://issues.redhat.com/browse/OCMUI-4623)
Addresses [OCMUI-4611](https://redhat.atlassian.net/browse/OCMUI-4611)

# How to Test

## Reproducing the bug (on `main`)

The page reload does **not** happen on every auto-refresh — only when a field inside the cluster's \`subscription\` object (which is part of the query key in \`useFetchCluster.ts\`) changes server-side. This happens roughly every ~5 minutes. When it does, React Query treats it as a new query, triggers a full re-fetch, and the changed data causes page reload.

Steps:
1. Open the browser DevTools Network tab
2. Navigate to a cluster's Machine Pools tab
3. Open the Add or Edit machine pool modal and begin filling in fields
4. **Wait ~5 minutes** (could be less) with the modal open — when the subscription data changes you will see network requests fire in the background and the modal may close or the form fields reset.

To accelerate the timer-based refresh (but note this alone does not trigger the bug — you still need to wait for the subscription field to change):
- Temporarily change \`longTimerSeconds\` from \`60\` to \`7\` in \`src/components/common/RefreshButton/RefreshButton.tsx\` so the refresh fires more frequently while you wait

## Verifying the fix (on this branch)

1. Repeat the same steps — open the modal and leave it open for several minutes
2. No cluster-details network requests should fire while the modal is open
3. The loading spinner on the cluster details header should not appear during the modal session
4. Form fields should remain stable regardless of how long the modal stays open

# Additional Information

This fix is also explicitly temporary — the correct long-term fix is tab-scoped refresh that removes this Redux dependency entirely

**Unit tests** were not added — the fix wires existing state into an existing Redux pattern already covered by \`ModalReducer.test.ts\` and \`MachinePools.test.jsx\`. The interaction between the timer and the modal guard is an integration concern better verified manually per the test steps above.

Assisted-by: Claude Code/claude-sonnet-4-6

# Review process

## QE Reviewer

[OCMUI-4623]: https://redhat.atlassian.net/browse/OCMUI-4623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OCMUI-4611]: https://redhat.atlassian.net/browse/OCMUI-4611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/MachinePools.jsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/MachinePools.jsx
@@ -153,7 +153,7 @@ const MachinePools = ({ cluster }) => {
     dispatch(openModal(modals.EDIT_MACHINE_POOL));
   };
 
-  const closeEditMachinePoolModal = () => {
+  const closeMachinePoolModal = () => {
     setEditMachinePoolId(undefined);
     setAddMachinePool(false);
     dispatch(closeModal());
@@ -365,7 +365,7 @@ const MachinePools = ({ cluster }) => {
           region={region}
           cluster={cluster}
           onSave={refreshMachinePools}
-          onClose={closeEditMachinePoolModal}
+          onClose={closeMachinePoolModal}
           isHypershift={isHypershift}
           machinePoolId={editMachinePoolId}
           machinePoolsResponse={machinePoolData}
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/MachinePools.jsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/MachinePools.jsx
@@ -75,10 +75,6 @@ const MachinePools = ({ cluster }) => {
   const isClusterAutoscalingModalOpen = useGlobalState((state) =>
     shouldShowModal(state, modals.EDIT_CLUSTER_AUTOSCALING_V2),
   );
-  const isEditMachinePoolModalOpen = useGlobalState((state) =>
-    shouldShowModal(state, modals.EDIT_MACHINE_POOL),
-  );
-
   const hasMachineConfiguration = useFeatureGate(ENABLE_MACHINE_CONFIGURATION);
   const organization = useGlobalState((state) => state.userProfile.organization);
 
@@ -145,17 +141,23 @@ const MachinePools = ({ cluster }) => {
 
   // Pause auto-refresh while the edit/add machine pool modal is open.
   // EditMachinePoolModal uses local state (not Redux), so anyModalOpen in ClusterDetails
-  // stays false and RefreshButton keeps firing. Syncing into the Redux modal slot fixes that.
-  // isEditMachinePoolModalOpen is intentionally excluded from deps to avoid a dispatch loop;
-  // it is read as a snapshot to guard against wiping unrelated modal state on mount/close.
-  React.useEffect(() => {
-    if (editMachinePoolId || addMachinePool) {
-      dispatch(openModal(modals.EDIT_MACHINE_POOL));
-    } else if (isEditMachinePoolModalOpen) {
-      dispatch(closeModal());
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [editMachinePoolId, addMachinePool]);
+  // stays false and RefreshButton keeps firing. We dispatch openModal/closeModal directly
+  // in the event handlers so no useEffect is needed.
+  const openEditMachinePoolModal = (machinePoolId) => {
+    setEditMachinePoolId(machinePoolId);
+    dispatch(openModal(modals.EDIT_MACHINE_POOL));
+  };
+
+  const openAddMachinePoolModal = () => {
+    setAddMachinePool(true);
+    dispatch(openModal(modals.EDIT_MACHINE_POOL));
+  };
+
+  const closeEditMachinePoolModal = () => {
+    setEditMachinePoolId(undefined);
+    setAddMachinePool(false);
+    dispatch(closeModal());
+  };
 
   React.useEffect(() => {
     if (!isDeleteMachinePoolPending && isDeleteMachinePoolSuccess && !isMachinePoolLoading) {
@@ -292,7 +294,7 @@ const MachinePools = ({ cluster }) => {
                         quotaReason
                       }
                       id="add-machine-pool"
-                      onClick={() => setAddMachinePool(true)}
+                      onClick={openAddMachinePoolModal}
                       variant={ButtonVariant.secondary}
                     >
                       Add machine pool
@@ -343,7 +345,7 @@ const MachinePools = ({ cluster }) => {
                   isDeleteMachinePoolPending={isDeleteMachinePoolPending}
                   isDeleteMachinePoolSuccess={isDeleteMachinePoolSuccess}
                   isDeleteMachinePoolError={isDeleteMachinePoolError}
-                  setEditMachinePoolId={setEditMachinePoolId}
+                  setEditMachinePoolId={openEditMachinePoolModal}
                   setHideDeleteMachinePoolError={setHideDeleteMachinePoolError}
                   cluster={cluster}
                   isMachinePoolError={isMachinePoolError}
@@ -363,10 +365,7 @@ const MachinePools = ({ cluster }) => {
           region={region}
           cluster={cluster}
           onSave={refreshMachinePools}
-          onClose={() => {
-            setEditMachinePoolId(undefined);
-            setAddMachinePool(false);
-          }}
+          onClose={closeEditMachinePoolModal}
           isHypershift={isHypershift}
           machinePoolId={editMachinePoolId}
           machinePoolsResponse={machinePoolData}
--- a/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/MachinePools.jsx
+++ b/src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/MachinePools.jsx
@@ -75,6 +75,9 @@ const MachinePools = ({ cluster }) => {
   const isClusterAutoscalingModalOpen = useGlobalState((state) =>
     shouldShowModal(state, modals.EDIT_CLUSTER_AUTOSCALING_V2),
   );
+  const isEditMachinePoolModalOpen = useGlobalState((state) =>
+    shouldShowModal(state, modals.EDIT_MACHINE_POOL),
+  );
 
   const hasMachineConfiguration = useFeatureGate(ENABLE_MACHINE_CONFIGURATION);
   const organization = useGlobalState((state) => state.userProfile.organization);
@@ -143,12 +146,15 @@ const MachinePools = ({ cluster }) => {
   // Pause auto-refresh while the edit/add machine pool modal is open.
   // EditMachinePoolModal uses local state (not Redux), so anyModalOpen in ClusterDetails
   // stays false and RefreshButton keeps firing. Syncing into the Redux modal slot fixes that.
+  // isEditMachinePoolModalOpen is intentionally excluded from deps to avoid a dispatch loop;
+  // it is read as a snapshot to guard against wiping unrelated modal state on mount/close.
   React.useEffect(() => {
     if (editMachinePoolId || addMachinePool) {
       dispatch(openModal(modals.EDIT_MACHINE_POOL));
-    } else {
+    } else if (isEditMachinePoolModalOpen) {
       dispatch(closeModal());
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [editMachinePoolId, addMachinePool]);
 
   React.useEffect(() => {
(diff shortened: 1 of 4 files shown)


src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/MachinePools.jsx
eliranmal, line 165: according to the explanation in the new code comment, i'd expect guards/conditions to be added around this `machinePoolOrNodePoolsRefetch()` call.

it seems more reasonable to prevent the data refresh rather than explicitly open or close the modal imperatively (i.e. not via user interaction).

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/MachinePools.jsx
```diff
@@ -140,6 +143,20 @@ const MachinePools = ({ cluster }) => {
     // eslint-disable-next-line  react-hooks/exhaustive-deps
   }, []);
 
+  // Pause auto-refresh while the edit/add machine pool modal is open.
+  // EditMachinePoolModal uses local state (not Redux), so anyModalOpen in ClusterDetails
+  // stays false and RefreshButton keeps firing. Syncing into the Redux modal slot fixes that.
+  // isEditMachinePoolModalOpen is intentionally excluded from deps to avoid a dispatch loop;
+  // it is read as a snapshot to guard against wiping unrelated modal state on mount/close.
+  React.useEffect(() => {
```
eliranmal: we have a guideline not to add `useEffect`s, as they increase complexity.

can we solve this problem another way?

eliranmal (commented):
see inline comments ✏️ 

additionally, the desc' says  _"the form fields reset mid-edit"_, but i've seen a different problem happening - the entire modal gets closed on data re-fetch.

it's also still unclear how to reliably reproduce the bug (the 'how to test' section ambiguously states that _"\[...] form fields may also reset mid-edit"_).  
AFAICT it's still sporadic in staging..  did you find a deterministic way to reproduce it?
without consistent repro', it's very hard to verify the problem was fixed.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/MachinePools.jsx
lizagilman: There is a section about it in the PR description

eliranmal: ok, i see an explanation is now added in the PR desc'  👌 

hmm it still has code-smell, but i think we can live with it as long as we keep close attention to the followup tasks. 
(i.e. open jira tickets with clear plan for a permanent solution, and mention them in code comments)

eliranmal: btw, the desc' mentions SDA-2249, but i think it's a mistake.  or at least i can't access this ticket in jira in order to verify..

lizagilman: yeah it's a comment on our code base , it's probably now https://redhat.atlassian.net/browse/OCMUI-594

eliranmal: for book-keeping, i'm posting the two related tickets that previously handled similar (identical?) issues:

- https://redhat.atlassian.net/browse/OCMUI-2568
- https://redhat.atlassian.net/browse/OCMUI-2478

eliranmal:
@lizagilman - after more functional testing, i see the data still refreshes (i.e. a spinner is seen behind the modal), on browser tab/window refocus.

this is a default behavior of react-query queries (per our config).  it can be controlled with the `refetchOnWindowFocus` flag.

as we're assuming here the bug is correlated to data-refreshes, i think this is a good use-case for switching of that flag.  WDYT?

---

p.s.  
there's also the related `refetchIntervalInBackground` and `refetchOnReconnect` you might want to explore (and possibly also `refetchOnMount`, but that seems to brittle to mess around with on this PR)

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/MachinePools.jsx
lizagilman: This one only re-fetches machine pool specific data after some crud operations, it's not the source of the auto-refresh problem, guarding against it won't solve the issue.

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/MachinePools.jsx
```diff
@@ -140,6 +140,17 @@ const MachinePools = ({ cluster }) => {
...
+  // Pause auto-refresh while the edit/add machine pool modal is open.
+  // EditMachinePoolModal uses local state (not Redux), so anyModalOpen in ClusterDetails
+  // stays false and RefreshButton keeps firing. Syncing into the Redux modal slot fixes that.
+  React.useEffect(() => {
+    if (editMachinePoolId || addMachinePool) {
+      dispatch(openModal(modals.EDIT_MACHINE_POOL));
+    } else {
+      dispatch(closeModal());
+    }
+  }, [editMachinePoolId, addMachinePool]);
```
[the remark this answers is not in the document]
lizagilman: dispatch is stable between renders
fixed the other issue

lizagilman:
> @lizagilman - after more functional testing, i see the data still refreshes (i.e. a spinner is seen behind the modal), on browser tab/window refocus.
> 
> this is a default behavior of react-query queries (per our config). it can be controlled with the `refetchOnWindowFocus` flag.
> 
> as we're assuming here the bug is correlated to data-refreshes, i think this is a good use-case for switching of that flag. WDYT?
> 
> p.s. there's also the related `refetchIntervalInBackground` and `refetchOnReconnect` you might want to explore (and possibly also `refetchOnMount`, but that seems to brittle to mess around with on this PR)

`useFetchMachineOrNodePools.ts` already has `refetchOnWindowFocus: !isModalOpen` . Not sure about other queries but anyway this is a little bit out of scope - the core issue is covered, no refresh when the modal is statically  opened.

eliranmal:
> the core issue is covered, no refresh when the modal is statically opened

@lizagilman i thought (as per the PR desc') the indication of a refresh is the spinner.  should the spinner still show?  do we only care about the firing of the specific query for `useFetchMachineOrNodePools`?

either way, i still can't reliably reproduce the bug in staging, so can't determine if the original problem was fixed (or if there was any problem, and/or what caused it).


https://github.com/user-attachments/assets/e61dba33-3345-4501-8006-52406ae1230a

dtaylor113 (approved):
Code LGTM -- clean, minimal fix that leverages the existing Redux modal infrastructure. The useEffect sync approach is a reasonable tradeoff for a hotfix, and the rationale for choosing it over call-site dispatching is well documented.

One note: the following local-state modals on cluster details are still unprotected from auto-refresh, worth tracking as follow-up:

- Billing Account (Overview tab) -- simpler form, most users would complete well within 60 seconds, unlikely to be impacted
- Channel Edit, Channel Group Edit (Upgrade Settings tab) -- also simple single-field forms, low risk
- Break Glass Credentials (Access Control tab → External Authentication) -- has 3 modals (view, create, delete), could involve more complex input
- External Auth Provider (Access Control tab → External Authentication) -- has 2 modals (edit, delete)
- Machine Pool Configuration (Machine Pools tab) -- not in production, behind enable-machine-configuration feature flag

dtaylor113 (approved):
Code looks even better me know with no useEffect().
Tested 3 rounds of navigate to cluster details, Add and Edit MP modals, change form, wait 2 minutes.
The modal remained open with no refresh.
Thanks @lizagilman 👍

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/MachinePools.jsx
```diff
@@ -140,6 +139,26 @@ const MachinePools = ({ cluster }) => {
...
+    setEditMachinePoolId(machinePoolId);
+    dispatch(openModal(modals.EDIT_MACHINE_POOL));
+  };
+
+  const openAddMachinePoolModal = () => {
+    setAddMachinePool(true);
+    dispatch(openModal(modals.EDIT_MACHINE_POOL));
+  };
+
+  const closeEditMachinePoolModal = () => {
```
eliranmal: nit:  this applies to both modal variants; please rename to `closeMachinePoolModal`

eliranmal (dismissed):
the bug still exists;  it seems that after some time (possibly after the first data refresh), when window/tab refocus occurs, the modal will close.  

see attached screen capture:

https://github.com/user-attachments/assets/471bbb13-6e87-4e20-b422-36c1977aadec

---

as `refetchOnWindowFocus` is already set for the node/machine-pools query (as  you mentioned), i suspect this happens at a higher level, at the rendering of the entire component tree.

maybe try declaring `refetchOnMount`?   tho, even that would probably not work - because this seems like a complete page reload..

**update:**  
i've tested locally, and  `refetchOnMount: false` indeed doesn't fix it 😞   
(not even `refetchOnReconnect: false`, or explicitly setting `refetchOnWindowFocus: false`)

shivanthi-amara (approved):
Tested the issue on multiple browsers, and I do not see any issue with the code changes. The form does not refresh, and the loading spinner on the cluster details header should not appear. LGTM

jcaianirh (approved):
I've tested this side by side with staging and also locally and observed that my edited machine pool modal will disappear on staging, however, with this fix, my changes are preserved and the bug no longer occurs.

src/queries/common/useFetchSubscription.ts
```diff
@@ -32,6 +32,7 @@ export const useFetchSubscription = (subscriptionID: string, mainQueryKey: strin
         isOSDCluster,
       };
     },
+    refetchOnWindowFocus: false,
```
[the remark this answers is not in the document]
lizagilman: out of scope, committed by mistake

src/queries/ClusterDetailsQueries/useFetchCluster.ts
```diff
@@ -29,6 +29,7 @@ export const useFetchCluster = (
       return response;
     },
     retry: false,
+    refetchOnWindowFocus: false,
```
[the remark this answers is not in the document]
lizagilman: out of scope, committed by mistake

src/queries/common/useFetchSubscription.ts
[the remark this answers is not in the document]
DcooperRH: This seems like a very valid suggestion here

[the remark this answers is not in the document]
DcooperRH: Nvm. I see this isn't part of the changes, was confused for a minute. Idk why but your comments above for this thread weren't showing here when I first saw this

eliranmal:
@lizagilman, after adding the changes from #568 into the branch of this PR (locally), i finally see a difference between staging and dev in my initial testing - dev' is working properly and the modal isn't getting closed (tested on ROSA classic)  👏 

as this was the only fix & scenario i could verify, i think it's in-scope for the fix, and we need to merge both changes (i.e. add `refetchOnWindowFocus: false` to useFetchCluster & useFetchSubscription, inside this PR).

or alternatively – if introducing only #568 will fix the issue, let's merge only it.

PR #568, opened 2026-06-01
fix: disable refetchOnWindowFocus for subscription and cluster queries

lizagilman:
Switching browser tabs triggered a window-focus refetch of useFetchSubscription. When any field in the subscription object had changed server-side, the new object produced a different query key in useFetchCluster, causing a cache miss and a full cluster re-fetch that reloaded the page. Disabling refetchOnWindowFocus on both queries stops the cascade; explicit invalidation via invalidateClusterDetailsQueries (timer and manual refresh) is unaffected.

# Summary

# Jira

# Additional information

# How to Test

# Screen Captures

|  |  |

# Review process

## QE Reviewer

### DIFF
--- a/src/queries/ClusterDetailsQueries/useFetchCluster.ts
+++ b/src/queries/ClusterDetailsQueries/useFetchCluster.ts
@@ -29,6 +29,7 @@ export const useFetchCluster = (
       return response;
     },
     retry: false,
+    refetchOnWindowFocus: false,
     enabled:
       !!subscription &&
       subscription.status !== SubscriptionCommonFieldsStatus.Deprovisioned &&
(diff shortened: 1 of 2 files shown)


src/queries/ClusterDetailsQueries/useFetchCluster.ts
```diff
@@ -29,6 +29,7 @@ export const useFetchCluster = (
       return response;
     },
     retry: false,
+    refetchOnWindowFocus: false,
```
eliranmal, line 32: it'd be less intrusive (to UX) to use `refetchOnWindowFocus: !isModalOpen` here instead

(as done in useFetchMachineOrNodePools)

PR #592, opened 2026-06-22
OCMUI-4669: Change the queryKey for useFetchCluster to prevent extra reloads of data

zherman0:
# Summary

Change the queryKey on useFetchCluster from the entire subscription object to just the subscription region.  This prevents reloading/caching of data anytime anything on the subscription object changes.  The only thing that will affect our query is the region id.

Also added a unit test to verify additional queries are not made when other parts of the subscription object change.

Update: This similar issue can be found on the upgrade wizard modal where data refreshes can cause the modal to close early.  I have therefore also removed the subscription object from the upgradeGateAgreements call.

AI help write the unit test: cursor/composer-fast-2.5 

# Jira

Fixes [OCMUI-4669](https://issues.redhat.com/browse/OCMUI-4669)

# How to Test

Functionally testing this has proven too difficult to easily reproduce.  Therefore, this test procedure focuses that the changes to the key values has been properly updated.

1. Before switching to this branch locally, use the TRQ tool in the lower right to find the react query cache.
2. Search using the filter for "fetchCluster".  Verify that the key shown has:
```
["fetchClusterDetails","clusterService","2r3j4lenv7hqt3sgnbss00c1li5ehgh5",{"billing_expiration_date":"0001-01-01T00:00:00Z","billing_marketplace_account":"767438687542","capabilities":[{"inherited":false,"name":"capability.cluster.manage_cluster_admin","value":"true"},...
```
3. Repeat for "upgradeGates" and verify key:
```
["fetchClusterDetails","fetchClusterDetails","upgradeGates","clusterService","2r3j4lenv7hqt3sgnbss00c1li5ehgh5",{"isAROCluster":false,"isOSDCluster":false,"isROSACluster":true,"subscription":{"billing_expiration_date":"0001-01-01T00:00:00Z","billing_marketplace_account":"767438687542","capabilities":...
```
3. Now load the PR branch and check the key values. They should be much cleaner and straightforward without the full subscription object:
fetchCluster
```
["fetchClusterDetails","fetchCluster","2r3j4lenv7hqt3sgnbss00c1li5ehgh5",null]
```
UpgradeGates
```
["fetchClusterDetails","clusterService","upgradeGates","2r3j4lenv7hqt3sgnbss00c1li5ehgh5",null]
```
The `null` values maybe different depending on how your cluster is created, but `null` represents the region id

For a final test, this cache is used for cluster details so verify that the cluster details pages work and that the refresh works for cluster details. 

# Review process

## QE Reviewer

[OCMUI-4669]: https://redhat.atlassian.net/browse/OCMUI-4669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/queries/ClusterDetailsQueries/useFetchCluster.test.ts
+++ b/src/queries/ClusterDetailsQueries/useFetchCluster.test.ts
@@ -3,7 +3,9 @@ import axios from 'axios';
 import { waitFor } from '@testing-library/react';
 
 import apiRequest from '~/services/apiRequest';
+import clusterService, * as clusterServiceModule from '~/services/clusterService';
 import { renderHook } from '~/testUtils';
+import { Subscription, SubscriptionCommonFieldsStatus } from '~/types/accounts_mgmt.v1';
 
 import { mockedCluster, mockSubscriptionData } from '../__mocks__/queryMockedData';
 
@@ -11,10 +13,23 @@ import { useFetchCluster } from './useFetchCluster';
 
 type MockedJest = jest.Mocked<typeof axios> & jest.Mock;
 const apiRequestMock = apiRequest as unknown as MockedJest;
+const mockGetClusterServiceForRegion = jest.spyOn(
+  clusterServiceModule,
+  'getClusterServiceForRegion',
+);
 
 const MAIN_QUERY_KEY = 'clusterDetails';
 
+const subscriptionWithRegion: Subscription = {
+  ...mockSubscriptionData,
+  rh_region_id: 'us-east-1',
+};
+
 describe('useFetchCluster hook', () => {
+  beforeEach(() => {
+    mockGetClusterServiceForRegion.mockReturnValue(clusterService);
+  });
+
   afterEach(() => {
     jest.clearAllMocks();
   });
@@ -73,5 +88,36 @@ describe('useFetchCluster hook', () => {
     expect(result.current.isError).toBe(true);
     expect(result.current.error?.name).toEqual(403);
     expect(result.current.error?.message).toEqual('Cluster does not exist');
+  });
+
+  it('does not refetch cluster details when subscription changes but rh_region_id stays the same', async () => {
+    const clusterID = 'mockedClusterID';
+    const isAROCluster = false;
+
+    apiRequestMock.get.mockResolvedValue({ data: mockedCluster });
+
+    const { result, rerender } = renderHook(
+      ({ subscription }: { subscription: Subscription }) =>
+        useFetchCluster(clusterID, subscription, isAROCluster, MAIN_QUERY_KEY),
+      { initialProps: { subscription: subscriptionWithRegion } },
+    );
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+    expect(apiRequest.get).toHaveBeenCalledTimes(1);
+
+    rerender({
+      subscription: {
+        ...subscriptionWithRegion,
+        status: SubscriptionCommonFieldsStatus.Disconnected,
+        display_name: 'Updated subscription',
+      },
+    });
+
+    await waitFor(() => {
+      expect(result.current.isFetching).toBe(false);
+    });
+    expect(apiRequest.get).toHaveBeenCalledTimes(1);
   });
 });
--- a/src/queries/ClusterDetailsQueries/useFetchCluster.test.ts
+++ b/src/queries/ClusterDetailsQueries/useFetchCluster.test.ts
@@ -1,15 +1,20 @@
+import React from 'react';
 import axios from 'axios';
 
-import { waitFor } from '@testing-library/react';
+import { QueryClientProvider } from '@tanstack/react-query';
+import { renderHook as rtlRenderHook, waitFor } from '@testing-library/react';
 
+import { queryClient } from '~/components/App/queryClient';
 import apiRequest from '~/services/apiRequest';
 import clusterService, * as clusterServiceModule from '~/services/clusterService';
 import { renderHook } from '~/testUtils';
 import { Subscription, SubscriptionCommonFieldsStatus } from '~/types/accounts_mgmt.v1';
 
 import { mockedCluster, mockSubscriptionData } from '../__mocks__/queryMockedData';
+import { queryConstants } from '../queriesConstants';
 
 import { useFetchCluster } from './useFetchCluster';
+import { invalidateClusterDetailsQueries } from './useFetchClusterDetails';
 
 type MockedJest = jest.Mocked<typeof axios> & jest.Mock;
 const apiRequestMock = apiRequest as unknown as MockedJest;
@@ -24,6 +29,9 @@ const subscriptionWithRegion: Subscription = {
   ...mockSubscriptionData,
   rh_region_id: 'us-east-1',
 };
+
+const queryClientWrapper = ({ children }: { children: React.ReactNode }) =>
+  React.createElement(QueryClientProvider, { client: queryClient }, children);
 
 describe('useFetchCluster hook', () => {
   beforeEach(() => {
@@ -90,6 +98,26 @@ describe('useFetchCluster hook', () => {
     expect(result.current.error?.message).toEqual('Cluster does not exist');
   });
 
+  it('fetches cluster details using default clusterService when rh_region_id is absent', async () => {
+    const clusterID = 'mockedClusterID';
+    const isAROCluster = false;
+
+    apiRequestMock.get.mockResolvedValueOnce({ data: mockedCluster });
+
+    const { result } = renderHook(() =>
+      useFetchCluster(clusterID, mockSubscriptionData, isAROCluster, MAIN_QUERY_KEY),
+    );
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(mockGetClusterServiceForRegion).not.toHaveBeenCalled();
+    expect(apiRequest.get).toHaveBeenCalledTimes(1);
+    expect(result.current.isError).toBe(false);
+    expect(result.current.data?.data.kind).toEqual(mockedCluster.kind);
+  });
+
   it('does not refetch cluster details when subscription changes but rh_region_id stays the same', async () => {
     const clusterID = 'mockedClusterID';
     const isAROCluster = false;
@@ -119,5 +147,35 @@ describe('useFetchCluster hook', () => {
       expect(result.current.isFetching).toBe(false);
     });
     expect(apiRequest.get).toHaveBeenCalledTimes(1);
+  });
+
+  it('is invalidated by invalidateClusterDetailsQueries', async () => {
+    const clusterID = 'mockedClusterID';
+    const isAROCluster = false;
+
+    queryClient.clear();
+    apiRequestMock.get.mockResolvedValue({ data: mockedCluster });
+
+    const { result } = rtlRenderHook(
+      () =>
+        useFetchCluster(
+          clusterID,
+          subscriptionWithRegion,
+          isAROCluster,
+          queryConstants.FETCH_CLUSTER_DETAILS_QUERY_KEY,
+        ),
+      { wrapper: queryClientWrapper },
+    );
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+    expect(apiRequest.get).toHaveBeenCalledTimes(1);
+
+    invalidateClusterDetailsQueries();
+
+    await waitFor(() => {
+      expect(apiRequest.get).toHaveBeenCalledTimes(2);
+    });
   });
 });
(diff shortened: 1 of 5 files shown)


src/queries/ClusterDetailsQueries/useFetchUpgradeGateAgreements.ts
```diff
@@ -21,12 +20,11 @@ export const useFetchUpgradeGateAgreements = (
...
   const { isLoading, data } = useQuery({
     queryKey: [
-      queryConstants.FETCH_CLUSTER_DETAILS_QUERY_KEY,
       mainQueryKey,
-      'upgradeGates',
       'clusterService',
+      'upgradeGates',
       clusterID,
-      subscription,
+      subscription?.subscription.rh_region_id,
```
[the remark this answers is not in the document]
zherman0: The enabled checks do not affect the data.  A cluster cannot change if it is ARO or managed.  As for status, archiving a cluster will not change the gate agreements data.  There is no issue with stale data.

src/queries/ClusterDetailsQueries/useFetchCluster.ts
```diff
@@ -18,7 +18,7 @@ export const useFetchCluster = (
   mainQueryKey: string,
 ) => {
   const { isLoading, data, isError, error, isFetching } = useQuery({
-    queryKey: [mainQueryKey, 'clusterService', clusterID, subscription],
+    queryKey: [mainQueryKey, 'fetchCluster', clusterID, subscription?.rh_region_id],
```
lizagilman: Why changing `clusterService` to `fetchCluster`?

zherman0: The 'clusterService' does not help identify the cache. I noticed that most of the current react queries use their purpose as the second key identifier.  I know we have a larger ticket to clean up these key names, but I thought I would at least make this one make more sense.  Looking at some of the other keys, i could leave "clusterService" and just add "fetchCluster" as well.

dtaylor113:
Hi @zherman0 , I having trouble verifying the changed queryKey/values in the "How To Test" section, it is difficult to find the exact matching key using dev tools Filter
CleanShot 2026-06-24 at 15 48 45

- Is there another way to verify "prevent extra reloads of data"?
- What other parts of the UI/cluster details should we test to see if any regressions by changing the query:
- Should we run the cluster upgrade w/ gates wizard to verify it still works? (if related)

zherman0:
> Hi @zherman0 , I verified the changed queryKey/values in the "How To Test" section, but how do I know this hasn't broken anything in the UI.
> 
>     * Should we run the cluster upgrade w/ gates wizard to verify it still works? (if related)
> 
>     * What other part of cluster details might be affected by this change?

@dtaylor113 
  I am still trying to come up with a better test.  The risk is super low as this only changes the key name.  The invalidate function works on a broad key so reset is not affected.
As for the wizard, I don't believe we need call useFetchcluster on the wizard; that would not make sense.
I think we just need to verify that the cluster details works and that refresh works.  I will add that to the test notes.

dtaylor113:
Hi @zherman0 , note, I edited the comment after you had replied.

zherman0:
> Hi @zherman0 , note, I edited the comment after you had replied.

@dtaylor113 -
Only the cache key name has changed for upgradeGate.  There should be no noticeable change.  There is no invalidation so that cannot be affect.

zherman0:
> Hi @zherman0 , I having trouble verifying the changed queryKey/values in the "How To Test" section, it is difficult to find the exact matching key using dev tools Filter 
>     * Is there another way to verify "prevent extra reloads of data"?
> 
>     * What other parts of the UI/cluster details should we test to see if any regressions by changing the query:
> 
>     * Should we run the cluster upgrade w/ gates wizard to verify it still works? (if related)

@dtaylor113 - In that image, what are you searching for?  The upgradeGates cache is at the top.

As for the other questions:
1) As I stated in the test section, I have not found a good way to test this.  In some ways, this is like changing a variable name and asking for a test to show it has not broken the code.
2) The query was not changed.  The name of the cache was.  A general test of cluster details should suffice
3) Again, only the name of the cache was changed. One could check that the upgrade code works if they wish.

src/queries/ClusterDetailsQueries/useFetchCluster.ts
lizagilman: Query keys changes can be sensitive so I'm a bit concerned about changing it on the fly. Have you checked if there are any references to this query key segment? They can silently break. It requires more attention and verification, and also I would expect a note on the PR explaining why is it safe and that it has been verified.. anyway, since we already have a ticket and anyway it's out of scope, I suggest to drop it from this PR.

lizagilman: But consider to do fix the same issue (with `subscription` object) in  https://github.com/RedHatInsights/uhc-portal/blob/main/src/queries/ClusterDetailsQueries/useFetchLimitedSupportReasons.ts#L21 and maybe in https://github.com/RedHatInsights/uhc-portal/blob/main/src/queries/ClusterDetailsQueries/useFetchInflightChecks.ts#L50

dtaylor113: Hi @zherman0, the rename from `'clusterService'` to `'fetchCluster'` is separate from the core fix and I'd agree with Liza's suggestion to drop it from this PR. Two components perform targeted invalidation using `'clusterService'` as a prefix match:

- [`OverviewBillingAccountModal.tsx` L64-71](https://github.com/RedHatInsights/uhc-portal/blob/main/src/components/clusters/ClusterDetailsMultiRegion/components/Overview/BillingAccount/OverviewBillingAccountModal.tsx#L64-L71)
- [`EditAutoNodeModal.tsx` L79-85](https://github.com/RedHatInsights/uhc-portal/blob/main/src/components/clusters/ClusterDetailsMultiRegion/components/Overview/EditAutoNodeModal/EditAutoNodeModal.tsx#L79-L85)

The rename would cause those targeted invalidation calls to silently stop matching. The auto-refresh timer covers it within ~60s, but it's still a subtle behavioral change that's unrelated to the OCMUI-4669 fix. `'clusterService'` is the established convention across sibling hooks in this directory — the rename could go into the existing cleanup ticket instead.

```suggestion
    queryKey: [mainQueryKey, 'clusterService', clusterID, subscription?.rh_region_id],
```

src/queries/ClusterDetailsQueries/useFetchCluster.test.ts
```diff
@@ -74,4 +89,35 @@ describe('useFetchCluster hook', () => {
...
+        ...subscriptionWithRegion,
+        status: SubscriptionCommonFieldsStatus.Disconnected,
+        display_name: 'Updated subscription',
+      },
+    });
+
+    await waitFor(() => {
+      expect(result.current.isFetching).toBe(false);
+    });
+    expect(apiRequest.get).toHaveBeenCalledTimes(1);
```
dtaylor113, line 149: Hi @zherman0, the `queryFn` has two branches — one for when `rh_region_id` is present (calls `getClusterServiceForRegion`) and one for when it's absent (falls back to default `clusterService`). The new test only covers the region-present path. There are at least 8 sibling hooks in this directory that explicitly test both the "with region" and "without region" paths (e.g., [`useDeleteHtpasswdUser.test.ts`](https://github.com/RedHatInsights/uhc-portal/blob/main/src/queries/ClusterDetailsQueries/AccessControlTab/UserQueries/useDeleteHtpasswdUser.test.ts#L18-L34)). Could you add a companion test for the `undefined` region case to stay consistent with that pattern?

```typescript
it('fetches cluster details using default clusterService when rh_region_id is undefined', async () => {
  const clusterID = 'mockedClusterID';
  const isAROCluster = false;

  apiRequestMock.get.mockResolvedValue({ data: mockedCluster });

  const { result } = renderHook(() =>
    useFetchCluster(clusterID, mockSubscriptionData, isAROCluster, MAIN_QUERY_KEY),
  );

  await waitFor(() => {
    expect(result.current.isLoading).toBe(false);
  });
  expect(apiRequest.get).toHaveBeenCalledTimes(1);
  expect(mockGetClusterServiceForRegion).not.toHaveBeenCalled();
});
```

src/queries/ClusterDetailsQueries/useFetchUpgradeGateAgreements.ts
dtaylor113, line 27: Hi @zherman0, this hook's query key is also being changed (replacing the full `subscription` object with `subscription?.subscription.rh_region_id`), but there's no test coverage for it. This hook runs at the page level — it's called in `useFetchClusterDetails` (line 74) and rendered persistently in `UpgradeAcknowledgeWarning` on the overview and upgrade settings tabs — so it's subject to the same unnecessary-refetch problem that `useFetchCluster` had.

The same "no refetch on subscription change" test you wrote for `useFetchCluster` would be valuable here. At minimum:
1. Subscription object changes with same `rh_region_id` → no refetch
2. Valid response returns gate agreements data

Would you be able to add a `useFetchUpgradeGateAgreements.test.ts`, or should we track it separately?

src/queries/ClusterDetailsQueries/useFetchCluster.test.ts
```diff
@@ -74,4 +89,35 @@ describe('useFetchCluster hook', () => {
     expect(result.current.error?.name).toEqual(403);
     expect(result.current.error?.message).toEqual('Cluster does not exist');
   });
+
+  it('does not refetch cluster details when subscription changes but rh_region_id stays the same', async () => {
```
dtaylor113, line 121: Hi @zherman0, since the query key structure is changing, it would be valuable to add a test confirming that `invalidateClusterDetailsQueries()` still correctly invalidates the `useFetchCluster` query with the new key. There's already a [predicate test in `useGetClusterDetails.test.ts`](https://github.com/RedHatInsights/uhc-portal/blob/main/src/queries/ClusterDetailsQueries/useGetClusterDetails.test.ts#L194-L200), but it tests the predicate in isolation — it doesn't exercise the actual hook's key. Something like:

```typescript
it('is invalidated by invalidateClusterDetailsQueries', async () => {
  const clusterID = 'mockedClusterID';
  const isAROCluster = false;

  apiRequestMock.get.mockResolvedValue({ data: mockedCluster });

  const { result } = renderHook(() =>
    useFetchCluster(clusterID, subscriptionWithRegion, isAROCluster, MAIN_QUERY_KEY),
  );

  await waitFor(() => {
    expect(result.current.isLoading).toBe(false);
  });
  expect(apiRequest.get).toHaveBeenCalledTimes(1);

  invalidateClusterDetailsQueries();

  await waitFor(() => {
    expect(apiRequest.get).toHaveBeenCalledTimes(2);
  });
});
```

This would catch any future key changes that accidentally break compatibility with the broad predicate-based invalidation.

src/queries/ClusterDetailsQueries/useFetchCluster.ts
zherman0: @dtaylor113 @lizagilman - To simplify this code, I have reverted the cache name back, even though at least 6 other services uses the same cache name. I know we have another ticket to deal with that issue.  But having the same queryKey for different queryFunctions is not good.

src/queries/ClusterDetailsQueries/useFetchCluster.test.ts
zherman0: Done

src/queries/ClusterDetailsQueries/useFetchUpgradeGateAgreements.ts
zherman0: Add test file and test cases for useFetchUpgradeGateAgreements @dtaylor113

src/queries/ClusterDetailsQueries/useFetchCluster.test.ts
zherman0: Ok, even though I change the key, I went ahead and added a test for this.

dtaylor113 (approved):
Hi @zherman0, yaml file LGTM.

**Spot-check verification — LGTM** ✅

Verified locally on ROSA HCP cluster:

1. **Cluster details overview loads correctly** — all fields populated (status, version, billing, network, nodes, etc.)
2. **Refresh/invalidation works** — enabled `delete_protection` via `ocm patch` CLI, clicked the Refresh button, and confirmed the overview updated to show Delete Protection as "Enabled". This exercises the full `refreshClusterDetails()` → predicate-based `invalidateClusterDetailsQueries()` → `useFetchCluster` refetch path with the new query key.
3. **Upgrade Settings tab loads correctly** — update status and available versions are displayed, confirming `useFetchUpgradeGateAgreements` fetches data successfully with the narrowed query key.

Code changes and new unit tests look good — see my earlier review comments which have been all addressed.

p.s. since the query key rename was reverted, pls remove "Test the billing account" &  "edit or autonode modal" —since those invalidation paths are unchanged

src/queries/ClusterDetailsQueries/useFetchCluster.ts
dtaylor113: ok, 👍

lizagilman (approved):

meirinetta (approved):
Tested locally, both key values (fetchCluster and upgradeGates) are now with subscription.rh_region_id field. Approving.

PR #599, opened 2026-06-24
OCMUI-4195: Fix infinite spinner for AI clusters with non-404 errors

lizagilman:
# Summary

Fix an infinite loading spinner on the cluster details page for AI (Assisted Installer) clusters that return a non-404 error from the assisted-installer API.

Unfortunately we don't have clear steps to reproduce the issue, and our mock data env cannot mock responses from Assisted-Installer service, so the only way to test is to tweak the code to return 403 error.  

Assisted by: Claude Code / claude-sonnet-4-6

# Jira

Fixes [OCMUI-4195](https://issues.redhat.com/browse/OCMUI-4195)

# Additional information

**Root cause:** `useFetchAiCluster` only caught 404 errors and re-threw everything else. Any other error (e.g. 500) propagated up to `useFetchClusterDetails`, which fell through to the `return { isLoading: true }` fallback — causing an infinite spinner with no error state.

**Fix 1 — `useFetchAiCluster.ts`:** Replaced the 404-only catch with a catch-all that always falls back to `fakeClusterFromSubscription`. This matches the existing intent for 404 and extends it to cover other transient or server errors.

**Fix 2 — `useFetchClusterDetails.ts`:** Destructure `isError` from `useFetchAiCluster` and include it in both the error-guard condition and the inner `isLoading` calculation, so an AI cluster error surfaces as an error state rather than a perpetual loading state.

# How to Test

1. To simulate the bug, temporarily modify `src/services/assistedService.ts` — replace these two lines:
   ```ts
   const getAIClustersBySubscription = ClustersAPI.listBySubscriptionIds;
   const getAICluster = ClustersAPI.get;
   ```
   with:
   ```ts
   const mock403 = (): Promise<any> => Promise.reject(
     Object.assign(new Error('Request failed with status code 403'),
       { isAxiosError: true, response: { status: 403, statusText: 'Forbidden',
         data: { code: 403, message: 'Unauthorized to access route (access review failed)' } } })
   );
   const getAIClustersBySubscription = (..._args: any[]): Promise<any> => mock403();
   const getAICluster = (..._args: any[]): Promise<any> => mock403();
   ```

2. Navigate to an AI (Assisted Installer) cluster's details page in the mock data env (https://prod.foo.redhat.com:1337/openshift/?env=mockdata).
3. **Before fix:** the cluster details page spins indefinitely with no error state.
4. **After fix:** the cluster details page renders an error state instead of spinning.

# Screen Captures

Before - infinite spinner

image

After - page renders correctly (with data from OCM only) 

image

# Review process

## QE Reviewer

[OCMUI-4195]: https://redhat.atlassian.net/browse/OCMUI-4195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/queries/ClusterDetailsQueries/useFetchAiCluster.ts
+++ b/src/queries/ClusterDetailsQueries/useFetchAiCluster.ts
@@ -23,14 +23,9 @@ const getAIClusterDetails = async (clusterID: string, subscription: Subscription
       cluster = fakeClusterFromAISubscription(subscription, aiCluster?.data || null);
       cluster.aiCluster = aiCluster.data;
     } catch (e) {
-      if (axios.isAxiosError(e) && e.response?.status === 404) {
-        // The cluster is garbage collected or the user does not have privileges
-        // eslint-disable-next-line no-console
-        console.info('Failed to query assisted-installer cluster id: ', clusterID);
-        cluster = fakeClusterFromSubscription(subscription);
-      } else {
-        throw e;
-      }
+      // eslint-disable-next-line no-console
+      console.info('Failed to query assisted-installer cluster id: ', clusterID);
+      cluster = fakeClusterFromSubscription(subscription);
     }
     try {
       const featureSupportLevels = await assistedService.getAIFeatureSupportLevels(
--- a/src/queries/ClusterDetailsQueries/useFetchAiCluster.ts
+++ b/src/queries/ClusterDetailsQueries/useFetchAiCluster.ts
@@ -49,11 +49,11 @@ const getAIClusterDetails = async (clusterID: string, subscription: Subscription
 };
 
 /**
- * Query responsible for fetching actions and permissions
- * @param subscriptionID subscription ID to pass into api call
+ * Query responsible for fetching AI cluster details
+ * @param clusterID cluster ID to pass into api call
  * @param mainQueryKey used for invalidation of the query (refetch)
- * @param subscriptionStatus status of the subscription for query enablement
- * @returns cloud providers array
+ * @param subscription subscription object used for query enablement and building fake cluster
+ * @returns AI cluster details or fake cluster assembled from subscription
  */
 export const useFetchAiCluster = (
   clusterID: string,
--- a/src/queries/ClusterDetailsQueries/useFetchAiCluster.test.ts
+++ b/src/queries/ClusterDetailsQueries/useFetchAiCluster.test.ts
@@ -1,0 +1,124 @@
+import { waitFor } from '@testing-library/react';
+
+import * as normalizeModule from '~/common/normalize';
+import { assistedService } from '~/services';
+import { renderHook } from '~/testUtils';
+import { Subscription, SubscriptionCommonFieldsStatus } from '~/types/accounts_mgmt.v1';
+
+import { useFetchAiCluster } from './useFetchAiCluster';
+
+jest.mock('~/common/normalize');
+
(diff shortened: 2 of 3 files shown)


src/queries/ClusterDetailsQueries/useFetchAiCluster.ts
```diff
@@ -23,14 +23,9 @@ const getAIClusterDetails = async (clusterID: string, subscription: Subscription
...
-      if (axios.isAxiosError(e) && e.response?.status === 404) {
-        // The cluster is garbage collected or the user does not have privileges
-        // eslint-disable-next-line no-console
-        console.info('Failed to query assisted-installer cluster id: ', clusterID);
-        cluster = fakeClusterFromSubscription(subscription);
-      } else {
-        throw e;
-      }
+      // eslint-disable-next-line no-console
+      console.info('Failed to query assisted-installer cluster id: ', clusterID);
```
dtaylor113, line 27: Hi @lizagilman — The JSDoc for `useFetchAiCluster` (line 52) was copied from `useFetchActionsPermissions.ts` in the same directory — it says "Query responsible for fetching actions and permissions", references params `subscriptionID` and `subscriptionStatus`, and claims `@returns cloud providers array`. None of that describes this hook.

Since you're modifying this function's error behavior, could you update the JSDoc to match? e.g.:
```ts
/**
 * Hook to fetch Assisted Installer cluster details
 * @param clusterID cluster ID to pass into the AI service call
 * @param mainQueryKey used for invalidation of the query (refetch)
 * @param subscription subscription object; query is disabled when undefined
 * @returns AI-enriched cluster data or fallback from subscription
 */
```
-thanks!

dtaylor113 (approved):
Hi, tested locally prior to fix (the infinite spinning) and after the fix (silent degradation).  

I'm still not 100% sure if silently obfuscating 404  ("No AI data exists for this cluster") and 403 ("You're not authorized to see AI data") errors is appropriate.  On the cluster list page we do trap and show similar errors, but that is a larger discussion beyond this PR.

I left one non-blocking request for a change.

lizagilman:
> Hi, tested locally prior to fix (the infinite spinning) and after the fix (silent degradation).
> 
> I'm still not 100% sure if silently obfuscating 404 ("No AI data exists for this cluster") and 403 ("You're not authorized to see AI data") errors is appropriate. On the cluster list page we do trap and show similar errors, but that is a larger discussion beyond this PR.
> 
> I left one non-blocking request for a change.

I agree, I was also thinking about it. We should probably have some kind of indication that something went wrong. Maybe some alert/message saying something like - "An error occurred. Showing partial results".  I will open a ticket.

src/queries/ClusterDetailsQueries/useFetchAiCluster.ts
lizagilman: Updated, thanks.

zherman0 (approved):
LGTM.  Tested locally and reviewed the code.  My issues have been addressed. Thanks

lizagilman:
Created [OCMUI-4731- Show user-facing alert when Assisted Installer data is unavailable on cluster details page ](https://redhat.atlassian.net/browse/OCMUI-4731) cc @dtaylor113

shivanthi-amara (approved):
Tested the changes, performed regression, and everything looks good so far. LGTM

PR #601, opened 2026-06-24
OCMUI-4069: Redesign Slack broken-links checker

DcooperRH:
# Summary

This PR changes the slack link checker workflow to show a summary overview of broken links. a summaryMode variable has been added to the script check-links.mjs. This mode outputs what is sent to slack as well as the github actions log it runs in. Flag --summary was added as well. You can `./run/check-links.mjs --summary` to see what output is sent to slack. 
To keep the message short, a max length of 12 broken links is added to the --summary output.

A thumbs down or thumbs up emoji is displayed if there were failures or no failures. Replaces the alarm and green check emojis

Also added conditional output so that if the github run fails, the table returns "❌ Link check failed to run" as the report

Assisted by cursor/composer-2-fast

# Jira

Fixes [OCMUI-4069](https://issues.redhat.com/browse/OCMUI-4069) 

# Additional information
Once this merges, I will need to adjust the current live slack link checker workflow so that the new variables are mapped correctly. 

# How to Test

There is not a straightforward way to test this. The good thing is that the code here really only affects the message sent to slack. Once it is merged to main, it can easily be tested in staging. 

You should check that the output from `./run/check-links.mjs ` is the same as it previously was 

If you really want to test this PR, it will require setting up a testing slack workflow and pushing this branch to the main branch of your own forked repository 

1. In your github fork, create a placeholder repository secret named SLACK_BROKEN_LINKS_TEST_WEBHOOK_URL `https://github.com/<username>/uhc-portal/settings/secrets/actions` THIS IS WHERE THE WEBHOOK URL WILL GO
2. Create a slack webhook. Slack --> Tools --> create a workflow --> from a webhook
3. Configure the webhook to send a dm to yourself (not the team channel), **note the webhook url**, paste it into your github secret 
4. example http body should look like 
```
{
  "report": "Example text",
  "run_url": "Example text",
}
```
5. The message body should look like this below 

Screenshot 2026-06-25 at 3 53 55 PM

6. Download this branch locally
7. create a file `.github/workflows/check-links-test.yml` copy the entire contents of `.github/workflows/check-links.yml` into this file but replace line 9 with name:  with `name: Check Links Test (Slack DM)`
8. In this testing file, replace the two usages of `webhook: ${{ secrets.SLACK_BROKEN_LINKS_WEBHOOK_URL }}` with `webhook: ${{ secrets.SLACK_BROKEN_LINKS_TEST_WEBHOOK_URL }}`
9. Push the branch to your repo, open a pr and merge into your main branch (**be really careful here that you don't accidentally push to upstream main**)
10. Now in your forked repo, the github action Check Links Test (Slack DM) should exist. Run it and check the results sent to you from your created webhook when it finishes running. 

After testing the PR, you would want to return your branches back to normal. Here are the steps to do that:

Fetch the latest state of the upstream/main branch - git fetch upstream main
Verify you are checked out on the main branch - git checkout main
Reset it to point to the same commit as upstream's main - git reset --hard upstream/main

Update your forked repo's main branch (origin/main) to the same commit - `git push origin main -f`
(MAKE SURE ORIGIN IS YOUR OWN FORK) `git remote -v `

# Screen Captures

| Link checker Before                                              | Link checker After                                   |
| ocmui4069_before |  Screenshot 2026-07-09 at 3 25 02 PM  |

Please note, the message will show as (edited) at the bottom, I think it has something to do with slack trying to unfurl the links, but there is not a way in the workflow builder to prevent link unfurling. We could wrap each link in backticks `` in the script, but the output would be less appealing imo.
# Review process

## QE Reviewer

[OCMUI-4069]: https://redhat.atlassian.net/browse/OCMUI-4069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


run/check-links.mjs
```diff
@@ -824,13 +862,7 @@ async function main() {
...
   displayResults(statusByUrl, redirectItems, verboseMode, redirectsMode);
 
-  const has404 = Object.values(statusByUrl).filter(
-    (result) => typeof result === 'number' && result === 404,
-  );
-  if (process.env.GITHUB_OUTPUT) {
-    fs.appendFileSync(process.env.GITHUB_OUTPUT, `has404=${has404.length > 0}\n`);
-    fs.appendFileSync(process.env.GITHUB_OUTPUT, `notFoundLength=${has404.length}\n`);
-  }
+  writeGithubActionOutputs(statusByUrl, redirectItems);
```
eliranmal: extracting this to a function is nice, but i think we need to rethink about the concept.

writing output to `GITHUB_OUTPUT` tightly-couples the link-checker script to external systems.

not sure how this change was accepted in the first place, but i think we should **go the other way**; removing dependency on this env-var, and just go back to only outputing the report to tty.

the slack app can then just show the output as-is.  it doesn't even have to know whether it's a successful output (no broken links) or not - that's just sugaring.  if we really want to show some emojis on success/failure, we can just include them in the original script output.

this loose-coupling will then translate in the future to much less cross-maintenance (e.g. every time the script changes, we need to change the slack app), and more stability (one side could break the other).  as a bonus, it'd also make the code more readable.

.github/workflows/check-links.yml
```diff
@@ -23,28 +23,31 @@ jobs:
...
+        if: steps.check-links.outcome == 'success'
         uses: slackapi/slack-github-action@v2.1.1
         with:
           webhook: ${{ secrets.SLACK_BROKEN_LINKS_WEBHOOK_URL }}
           webhook-type: webhook-trigger
           payload: |
-            has404: "${{ steps.check-links.outputs.has404 }}"
-            404_length: "${{ steps.check-links.outputs.notFoundLength }}"
-            message: "Broken links found"
+            report: ${{ toJSON(steps.check-links.outputs.report) }}
```
dtaylor113, line 42: The summary table in `displaySummaryTable()` uses `padStart`/space-padding for column alignment — this is designed for a monospace font (terminal). When Slack renders the `report` text in its default proportional font, the Category and Count columns won't line up.

On the Slack Workflow Builder side, wrapping the `{{report}}` variable in triple backticks (`` ``` ``) would render it as a code block with a monospace font, preserving the column alignment for free:

image

.github/workflows/check-links-test.yml
```diff
@@ -0,0 +1,66 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: '24'
```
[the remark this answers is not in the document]
DcooperRH: This matches existing repo convention

run/check-links.mjs
```diff
@@ -663,29 +662,57 @@ function displayUsageNotes(verbose) {
...
-  console.log(`Server errors (5xx)               ${formatCount(serverErrors.length, true)}`);
-  console.log(`Request errors                    ${formatCount(errors.length, true)}`);
-  console.log('---------------------------------- ------');
-  console.log(`Total URLs checked                ${formatCount(totalChecked)}`);
+  if (process.env.GITHUB_ACTIONS) {
+    console.log('---LINK_CHECK_SLACK_SUMMARY---');
+    console.log(summary);
+    console.log('---END_LINK_CHECK_SLACK_SUMMARY---');
+    return;
+  }
```
dtaylor113: Consider writing directly to `$GITHUB_OUTPUT` instead of scraping stdout with sentinel markers.

`formatSummaryReport()` already returns the exact string you need — you can write it to `$GITHUB_OUTPUT` directly from Node, which is the standard GitHub Actions pattern and avoids the `tee` → `sed` strip → `sed` scrape pipeline in the workflow:

```suggestion
  if (process.env.GITHUB_OUTPUT) {
    const fs = await import('fs');
    const delimiter = 'EOF_REPORT';
    fs.appendFileSync(
      process.env.GITHUB_OUTPUT,
      `report<<${delimiter}\n${summary}\n${delimiter}\n`
    );
  }
```

This eliminates the sentinel markers, the ANSI stripping (`sed`), the intermediate files (`report.txt`, `report-clean.txt`, `slack-report.txt`), and the implicit coupling between the script's stdout format and the workflow's shell scraping. The workflow step simplifies to just `run: node run/check-links.mjs`.

.github/workflows/check-links.yml
```diff
@@ -23,28 +23,32 @@ jobs:
...
-        run: node run/check-links.mjs
+        run: |
+          node run/check-links.mjs 2>&1 | tee report.txt
+          sed 's/\x1b\[[0-9;]*m//g' report.txt > report-clean.txt
+          sed -n '/---LINK_CHECK_SLACK_SUMMARY---/,/---END_LINK_CHECK_SLACK_SUMMARY---/p' report-clean.txt | sed '1d;$d' > slack-report.txt
+          {
+            echo 'report<<EOF_REPORT'
+            cat slack-report.txt
+            echo 'EOF_REPORT'
+          } >> "$GITHUB_OUTPUT"
```
dtaylor113, line 33: Companion to the [suggestion in `check-links.mjs`](https://github.com/RedHatInsights/uhc-portal/pull/601#pullrequestreview-4575056224) — if the script writes the `report` output directly to `$GITHUB_OUTPUT`, this entire pipeline (`tee` → `sed` ANSI strip → `sed` sentinel scrape → heredoc) collapses to a single line:

```suggestion
        run: node run/check-links.mjs
```

No intermediate files, no implicit coupling between marker strings in the script and `sed` patterns here.

dtaylor113 (changes requested):

.github/workflows/check-links.yml
DcooperRH: Nice, thanks for the suggestion, was able to implement the same 

Screenshot 2026-06-25 at 3 50 48 PM

run/check-links.mjs
DcooperRH: Made the change here. This conflict's with Lyn's point above regarding removing process.env.GITHUB_OUTPUT all together. I believe the changes that have been made now do de-couple the script and slack output quite a bit more than when Lyn originally reviewed. Wdyt @eliranmal ?

t-hendricks (approved):
Tested the `check-links-test.yml` locally. Output lgtm! 
@DcooperRH Is there a plan to remove the test file?

DcooperRH:
> Tested the `check-links-test.yml` locally. Output lgtm! @DcooperRH Is there a plan to remove the test file?

Thanks @t-hendricks , I just removed it

run/check-links.mjs
eliranmal: IMO writing to `GITHUB_OUTPUT` is no different than writing to `GITHUB_ACTIONS`.  the script is still strongly coupled to an external system.

my previous feedback still holds.

eliranmal: if the script already returns the exact string we need, we should just read it in the GH workflow.

if the problem is the output is too large (i.e. returns more than just the summary), we need to **adapt the script CLI**.  
the most straight-forward way is to introduce an option to allow only outputting the summary, e.g. `--summary`, and pass it from the workflow call.

.github/workflows/check-links.yml
DcooperRH: Added this change

run/check-links.mjs
DcooperRH: Thanks @eliranmal , I've made these changes below 

- Removed all GITHUB_OUTPUT usage from check-links.mjs
- Added a --summary flag that prints only the summary table to stdout. The full error details go to stderr so the Actions log remains the same
- The workflow captures stdout and sets the step output from there

.github/workflows/check-links.yml
DcooperRH: Actually, due to Lyn's comments above on the coupling concers, this is now changed to: 
```
       run: |
          report=$(node run/check-links.mjs --summary)
          {
            echo "report<<EOF"
            echo "$report"
            echo "EOF"
          } >> "$GITHUB_OUTPUT"
```

We can't just use `run: node run/check-links.mjs` now unless the script writes to $GITHUB_OUTPUT

run/check-links.mjs
eliranmal: amazing!  thanx

run/check-links.mjs
```diff
@@ -27,6 +26,20 @@ const args = process.argv.slice(2);
...
+const summaryMode = args.includes('--summary');
+
+/**
+ * Writes informational output. In --summary mode, routes to stderr so stdout
+ * can be reserved for the summary table only.
+ * @param {...unknown} args - Values to print
+ */
+function log(...messageArgs) {
+  if (summaryMode) {
+    console.error(...messageArgs);
```
eliranmal: there's no reason why summary-mode needs to output to stderr;  summary does not indicate an error status.

see my [other comment][1] for a suggestion that'll help get rid of this.


[1]: https://github.com/RedHatInsights/uhc-portal/pull/601/changes#r3513182400

run/check-links.mjs
```diff
@@ -652,40 +666,72 @@ function displayGrandTotal(grandTotal) {
...
-  console.log(`Total URLs skipped                ${formatCount(skipped.length)}`);
-  console.log(`Success                           ${formatCount(success.length)}`);
-  console.log(`Redirects                         ${formatCount(redirects.length)}`);
-  console.log(`Redirects errors                  ${formatCount(redirectErrorCount, true)}`);
-  console.log(`Client errors (4xx)               ${formatCount(clientErrors.length, true)}`);
-  console.log(`Server errors (5xx)               ${formatCount(serverErrors.length, true)}`);
-  console.log(`Request errors                    ${formatCount(errors.length, true)}`);
-  console.log('---------------------------------- ------');
-  console.log(`Total URLs checked                ${formatCount(totalChecked)}`);
+  if (summaryMode) {
```
eliranmal: the conditioning should be the other way around; the summary section always gets printed, and only if summary-mode is off everything else gets printed.

eliranmal (changes requested):

run/check-links.mjs
DcooperRH: The issue I am having with this is keeping the full details in the github actions log. The stderr usage was a bit of a hack to with the goal of keep summary on stdout for Slack capture, but print detail sections to stderr in `--summary` mode so the log stays useful without coupling the script to $GITHUB_OUTPUT

eliranmal: why not just execute the script twice from the GH action?   (once with `--summary` and once without)

eliranmal: also, why do we even need the full details in the GH actions log?  the purpose of this workflow is just to report the summary.

if we want to look at the details, we can run the script locally.

DcooperRH: We could execute the script twice.. it just seems a little resource intensive. Also I'm concerned about hitting rate limits and getting our requests blocked if we are hitting each link twice.

Regarding having the full details, if slack gets pinged showing link errors, it is nice to be able to click the run url to see exactly which link(s) are broken, rather than running again in your terminal.

.github/workflows/check-links.yml
dtaylor113: Hi @DcooperRH, so is the Screenshot's After image correct?  Trying to determine how this affects the slack message.

DcooperRH: @dtaylor113 Yes the After image is still correct

DcooperRH: Here is my latest run
Screenshot 2026-07-06 at 1 20 05 PM

I believe the errors count is because of rate limit issues from running it so much

DcooperRH:
I will delete check-links-test.yml before merging

dtaylor113 (approved):
yml and mjs LGTM, hard to determine if check-links runs same as before as both runs currently result in a lot of 404s due to bandwidth limits(?)

run/check-links.mjs
eliranmal: hmm good point..  we wouldn't wanna duplicate all network calls.

hmm i don't know.  i still don't like the idea of using stderr not for its intended purpose. 

if all we need is to see the broken links easily, i think it'd be a good idea to just include (only) the broken links in the summary, if there are any.  
WDYT?

that way we don't have to run twice, and still get a more useful report.

DcooperRH: There's no more usages of stderr in the code. I worry about the message getting too busy if we are sending all broken links and redirect errors.

Edit: hold on I forgot we have to add it back to get the full report in the actions log

DcooperRH: We can do it that way, just show the broken links. 
I guess we have to determine the trade off of 

Option 1) Implementing it with stderr, which would report the summary to slack and the full run/check-links output to the github actions logs

Option 2) Just show only the broken links in the slack message, no summary table. 

I'm indifferent to the two, there's pros and cons to both. Do you have any input on this @dtaylor113 ?

eliranmal: > Option 1) Implementing it with stderr, which would report the summary to slack and the full run/check-links output to the github actions logs

i don't think misusing stderr should be an option.  once we do that, we're breaking the basic way shell scripts function.  what happens when the script outputs an actual error (which naturally goes to stderr)?

> Option 2) Just show only the broken links in the slack message, no summary table.

why is it either/or?  there's no reason why we can't include both.

i think both the summary table and the broken links are useful info, and under normal circumstances the broken links list will never be so long as to hinder the usability of that message.

DcooperRH: You're right it doesn't have to be either or. I was misreading. 
What do you think about this approach, the script now writes the summary to /tmp/link-check-summary.txt (--report-file) and is sent to slack that also lists the broken links. While the actions log still holds the full output. 

Screenshot 2026-07-07 at 11 53 55 AM


this is the github actions run --> https://github.com/DcooperRH/uhc-portal/actions/runs/28886960405/job/85689522687

DcooperRH: We could also just drop the Broken links: section in the slack message and the report-file will just write the summary while the github actions log contains the full script output

DcooperRH: Here it is formatted a bit better. 

Screenshot 2026-07-07 at 3 48 22 PM

eliranmal: why do we need an intermediary file?

i also still struggle to see the benefit of storing the full log if all useful info is already provided.

let me try and make my proposal clearer:

- GH action calls the script once
- GH action passes the `--summary` flag to the script
- script (in `--summary` mode) outputs the summary-table **and** any broken links found
- GH action relays that text as-is to slack (no write-to-file needed, the action can store the script output on the action context for the next step to read)
- slack presents that summarized info for users (summary table and broken links).  users don't need to visit the GH actions page to get details

eliranmal: > We could also [...] just write the summary while the github actions log contains the full script output

but to do that you'd need **two** outputs again – and we've already ruled out that option, cuz you'd have to either run the script twice, or misuse stderr.

DcooperRH: Sorry by full script output I just meant listing only the broken links, just the section "Broken links:" in the above screenshot in the github log, but I like your proposal above

DcooperRH: @eliranmal I believe the changes you requested have been added. 
> GH action passes the --summary flag to the script

By the above you just mean passing --summary in the check-links.yml run right?

eliranmal: ok, i'll have another look today.

and yes - that's what i meant.

run/check-links.mjs
```diff
@@ -61,6 +61,7 @@ Options:
   -v, --verbose  Show detailed URL listings for all categories
                  (By default, only error URLs are displayed)
   -r, --redirects Show ONLY redirected URLs with their redirect targets
+  --summary      Print summary table and broken links only (for CI / Slack)
```
eliranmal: nit:

no need to add "for CI / Slack" (separation of concerns!  the script should be agnostic to whomever consumes it).

actually this flag is gonna be hella useful in local runs, too :)

.github/workflows/check-links-test.yml
eliranmal, line 1: why do we need a "test" workflow?

we can't test changes to GH actions before merge to upstream/main either way.

run/check-links.mjs
```diff
@@ -311,7 +312,9 @@ async function testRedirectUrls(statusByUrl) {
   }
 
   // Test redirect destinations
-  console.log('\nTesting redirect destinations...');
+  if (!summaryMode) {
+    console.log('\nTesting redirect destinations...');
+  }
```
eliranmal, line 317: this seems redundant.  what "redirect destinations"?

(we agreed to steer away from the approach of misusing stderr, and this seems like a residue of that method..)

.github/workflows/check-links-test.yml
DcooperRH: This is just here as I've been running a test github action from my forked pr to test the slack output. I'm deleting this file before merging

run/check-links.mjs
DcooperRH: I don't think this is stderr related. This is existing from the original script, I've just excluded it from the summary mode. 
The redirect destinations are the URLs we fetch after detecting 3xx redirects. So this log statement shows right before the progress bar is shown when running the regular script, no summary mode `run/check-links.mjs`
We can remove it and just rely on the progress bar

.github/workflows/check-links-test.yml
eliranmal: why need a separate file, what's different about it?   does the original workflow not allow running from your fork?

note that running GH actions from forks _could be a security risk_.  

(this is the reason, for example, we ditched any use of the `pull_request_target` event types, and why our pipelines have been suspended for weeks while we made that change..)

run/check-links.mjs
eliranmal: ohhh i see i see - "redirect" in terms of "link redirects".  got ya 👌 

we can keep it, it don't matter.

.github/workflows/check-links-test.yml
eliranmal: generally, it would have been better to just do whatever change you need in the original file - if you needed it for testing, and then revert just that change later.  

having 2 files is very confusing to reviewers (i have to keep track of the differences between the two files and between them and the original, instead of git comparing changes for me..).

DcooperRH: I should have cleared up the two file thing more in the PR description. I left a comment early in the PR that I would delete the test file before merging. 

Good point on the security risk, check-links-test.yml was only for manual Slack testing on my fork. It has a separate test webhook and fork-only secret. I’m deleting it before merging.
check-links.yml only runs on upstream via schedule and workflow_dispatch. There’s no pull_request trigger, and I’m not introducing pull_request_target.

DcooperRH:
Hi @dtaylor113 and @t-hendricks , thank you for the approvals earlier. This pr has changed quite a bit since you originally approved so if you could look it over again that would be great. Thanks

PR #608, opened 2026-06-30
OCMUI-4743: Implement confirmation modal for updating machine pools

zherman0:
# Summary

Added a confirmation modal that prompts users before proceeding with the update of machine pools. The modal includes options to confirm or cancel the update, ensuring users have a chance to review their action. Updated tests to verify the modal behavior and ensure no updates occur if the confirmation is canceled.

NOTE: There are 12 files updated, but only 1 with the code changes and the other 11 are mock test files.

Assisted by: Cursor/composer-fast-2.5

# Jira

Fiixes [OCMUI-4743](https://issues.redhat.com/browse/OCMUI-4743)

# How to Test

## This is how to do it with a real cluster, but there is a current bug in the upgrade system that is preventing upgrades currently (OCPBUGS-67258).  Therefore, please use the mock server steps.
1. Create HCP ROSA cluster with multiple machine pools that are back leveled
2. Go to the Settings tab and manual update the control plane to the latest version.  After that completes, the machine pools will be back leveled from the control plane.
3. Now go to Cluster details -> Machine pools
4. There should be a alert banner that says you can update all machine pools. Click the link
5. A confirmation modal should appear.  Click "Update machine pools"
6. Machine pools should update

## Mock server

1. Go to the mock server [https://prod.foo.redhat.com:1337/openshift/clusters/list?env=mockdata](https://prod.foo.redhat.com:1337/openshift/clusters/list?env=mockdata)
2. Find cluster "hpyershift-ready" and click for details
3. Go to "Machine pools" tab
4. There should be alert with "Update all machine pools now"
5. Click on that and you will see the confirmation modal
6. Click cancel will close the window
7. Click on "Update machine pools" and look at the XHR of devtools.  You will see the update calls being made even though they will not work since it is mock.

# Screen Captures
Screenshot From 2026-06-30 12-48-16

# Review process

## QE Reviewer

[OCMUI-4743]: https://redhat.atlassian.net/browse/OCMUI-4743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/UpdateMachinePools/UpdateAllMachinePools.tsx
```diff
@@ -71,16 +74,26 @@ const UpdateAllMachinePools = ({
...
       controlPlaneRawVersion,
       region,
     );
     setPending(false);
-    setErrors(errors);
+    setErrors(updateErrors);
+
+    if (updateErrors.length === 0) {
+      setIsConfirmModalOpen(false);
+    }
```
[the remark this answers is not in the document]
zherman0: Resolved

jmekkatt (changes requested):
I came across a scenario during testing. When the Update Machine Pools modal was open, I followed the normal workflow and clicked Update Machine Pools. The loading spinner kept running for a long time, and I later realized that my network connection had been lost.
Interestingly, when I tried to close the modal by clicking the Close button, it didn't close and remained open. It looks like something is preventing the user from closing the modal while the request is in progress. The "Cancel" button also disabled from the model.
I also simulated this behavior using the browser DevTools by throttling/disconnecting the network, and I was able to reproduce the same issue there as well.

https://github.com/user-attachments/assets/d245c95b-70e4-4d6d-a0f0-32c2f51a9ebb



Could you please check @zherman0 ?

zherman0:
@jmekkatt - Fixed this issue.  I was waiting for the API to return before allowing the close which makes no sense in the situation.

hcpcluster.json.json
t-hendricks, line 1: You may want to remove this file `hcpcluster.json.json.`

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/UpdateMachinePools/UpdateAllMachinePools.tsx
```diff
@@ -143,6 +154,22 @@ const UpdateAllMachinePools = ({
...
+        <Modal
+          modalSize="small"
+          title="Update machine pools"
+          onClose={closeConfirmModal}
+          primaryText="Update machine pools"
+          secondaryText="Cancel"
+          onPrimaryClick={updateNodePools}
+          onSecondaryClick={closeConfirmModal}
+          isPrimaryDisabled={pending}
+        >
```
t-hendricks, line 167: The secondary `Cancel` button should be disabled while a fetch request is in flight. Having the button enabled can cause confusion and lead users to think that API requests can be cancelled while in the middle of a request. 
In @jmekkatt's case, this may require proper network-error handling with either a `try/catch` or a `try/finally` block, and possibly an explicit network-error message.

t-hendricks (commented):
After clicking "**_Update machine pools_**", the modal stays open until the API returns (unless the user dismisses it). During that time, there is no in-modal spinner.

https://github.com/user-attachments/assets/ae304fc0-1dcf-471f-bde2-43c02b1dc39b

src/components/clusters/ClusterDetailsMultiRegion/components/MachinePools/UpdateMachinePools/UpdateAllMachinePools.tsx
zherman0: Since the error handling exist at the parent and not in the modal, the right solution here is to close the modal as soon as either the update or cancel are pressed.  Then I do not need to worry about timing issues or disable buttons that have no use.  Thanks for the help @t-hendricks

t-hendricks: Accepted. Resolving this thread.

DcooperRH:
nit here, but I'm wondering if the "now" in 'Update all machine pools now' was intentional as clicking the link would update them right away. I always thought the 'now' looked funny tbh, but sorta made sense because you'd click it and they'd update right then. 
Now that there is a confirmation modal, it might be good to change the link to just "Update all machine pools"

zherman0:
> nit here, but I'm wondering if the "now" in 'Update all machine pools now' was intentional as clicking the link would update them right away. I always thought the 'now' looked funny tbh, but sorta made sense because you'd click it and they'd update right then. Now that there is a confirmation modal, it might be good to change the link to just "Update all machine pools"

@DcooperRH - That is a fair point.  I think the "now" did not mean click and it would do it immediately as much as it was referring to the normal upgrade path that has manual(now) versus automatic updates (future).  There, when you upgrade the control plane, the manual means it would do it "now" versus scheduling for the the future and I believe that is the messaging that is being conveyed.  

I think I would like to keep it with "now" so there is no confusion that we cannot schedule it for the future (which is not possible).  Let me know what you think and we can take it back to UX if you still have concerns.

DcooperRH:
> @DcooperRH - That is a fair point. I think the "now" did not mean click and it would do it immediately as much as it was referring to the normal upgrade path that has manual(now) versus automatic updates (future). There, when you upgrade the control plane, the manual means it would do it "now" versus scheduling for the the future and I believe that is the messaging that is being conveyed.
> 
> I think I would like to keep it with "now" so there is no confusion that we cannot schedule it for the future (which is not possible). Let me know what you think and we can take it back to UX if you still have concerns.

Got it. That makes more sense

